PR #8, opened 2021-04-11
Adds charging, plugged in, and low fuel sensors

badguy99:
Adds binary sensors for:
* battery charging
* car plugged in
* low fuel


fuatakgun (approved):
Looks good, thanks

PR #19, opened 2021-04-12
use Tz data so last update time shows correctly

badguy99:
Time data seems to come in Western European time -
image

This was previously causing the last update time to show as "In 30 minutes" for example, for me in the UK, as the time often looked like it was in the future.

This adds Tz data to the last update time, so the system can work out local time and display this
image
image
image

### DIFF
--- a/custom_components/kia_uvo/Vehicle.py
+++ b/custom_components/kia_uvo/Vehicle.py
@@ -1,5 +1,6 @@
 import logging
 
+import pytz
 import re
 import requests
 from datetime import datetime, timezone
@@ -83,11 +84,12 @@ class Vehicle(object):
         self.token = self.kia_uvo_api.login()
 
     def set_last_updated(self):
+        kia_tz = pytz.timezone("Europe/Paris")
         m = re.match(
             r"(\d{4})(\d{2})(\d{2})(\d{2})(\d{2})(\d{2})",
             self.vehicle_data["vehicleStatus"]["time"],
         )
-        self.last_updated = datetime(
+        _last_updated = datetime(
             year = int(m.group(1)),
             month = int(m.group(2)),
             day = int(m.group(3)),
@@ -95,6 +97,7 @@ class Vehicle(object):
             minute = int(m.group(5)),
             second = int(m.group(6)),
         )
+        self.last_updated = kia_tz.localize(_last_updated)
     
     def set_engine_type(self):
         if "dte" in self.vehicle_data["vehicleStatus"]:
--- a/custom_components/kia_uvo/Vehicle.py
+++ b/custom_components/kia_uvo/Vehicle.py
@@ -84,7 +84,6 @@ class Vehicle(object):
         self.token = self.kia_uvo_api.login()
 
     def set_last_updated(self):
-        kia_tz = pytz.timezone("Europe/Paris")
         m = re.match(
             r"(\d{4})(\d{2})(\d{2})(\d{2})(\d{2})(\d{2})",
             self.vehicle_data["vehicleStatus"]["time"],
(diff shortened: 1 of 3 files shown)


badguy99:
Don't merge this yet. I need to sort a fix for the update code -

```
 2021-04-12 22:57:48 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved
Traceback (most recent call last):
  File "/config/custom_components/kia_uvo/__init__.py", line 83, in update
    _LOGGER.debug(f"{DOMAIN} - Check last update of vehicle {vehicle.last_updated} {datetime.now()} {FORCE_SCAN_INTERVAL} {datetime.now() - vehicle.last_updated > FORCE_SCAN_INTERVAL}")
TypeError: can't subtract offset-naive and offset-aware datetimes
```

Will take a look at this later today

fuatakgun:
because of this comparison, i had removed timezone from last updated.

you can fix the regex expression like this and you can use `datetime.utcnow()` for comparisons

```
m = re.match(r"(\d{4})(\d{2})(\d{2})(\d{2})(\d{2})(\d{2})", status["lastStatusDate"])
        time = datetime(
            year=int(m.group(1)),
            month=int(m.group(2)),
            day=int(m.group(3)),
            hour=int(m.group(4)),
            minute=int(m.group(5)),
            second=int(m.group(6)),
            tzinfo=timezone.utc
        )
```

custom_components/kia_uvo/__init__.py
```diff
@@ -80,9 +80,9 @@ async def update(event_time):
         await vehicle.update()
         if (event_time.hour < NO_FORCE_SCAN_HOUR_START and event_time.hour >= NO_FORCE_SCAN_HOUR_FINISH):
             _LOGGER.debug(f"{DOMAIN} - We are in force hour zone {event_time}")
-            _LOGGER.debug(f"{DOMAIN} - Check last update of vehicle {vehicle.last_updated} {datetime.now()} {FORCE_SCAN_INTERVAL} {datetime.now() - vehicle.last_updated > FORCE_SCAN_INTERVAL}")
+            _LOGGER.debug(f"{DOMAIN} - Check last update of vehicle {vehicle.last_updated} {datetime.now(timezone.utc)} {FORCE_SCAN_INTERVAL} {datetime.now(timezone.utc) - vehicle.last_updated > FORCE_SCAN_INTERVAL}")
```
fuatakgun, line 83: this might be simpler `datetime.utcnow()`

custom_components/kia_uvo/Vehicle.py
```diff
@@ -83,18 +84,20 @@ def login(self):
         self.token = self.kia_uvo_api.login()
 
     def set_last_updated(self):
+        kia_tz = pytz.timezone("Europe/Paris")
```
fuatakgun, line 87: can we use constants here?

custom_components/kia_uvo/__init__.py
badguy99: that gives a native datetime rather than an aware one - I had to double check this too, so doesn't do the subtraction.

custom_components/kia_uvo/Vehicle.py
badguy99: Updated to use constants. Also now normalizing to use UTC for both side of the subtraction, so it isn't affected by daylight savings offsets changing for timezone time, but not UTC.

badguy99:
> m = re.match(r"(\d{4})(\d{2})(\d{2})(\d{2})(\d{2})(\d{2})", status["lastStatusDate"])
>         time = datetime(
>             year=int(m.group(1)),
>             month=int(m.group(2)),
>             day=int(m.group(3)),
>             hour=int(m.group(4)),
>             minute=int(m.group(5)),
>             second=int(m.group(6)),
>             tzinfo=timezone.utc
>         )

It didn't see right to add `tzinfo=timezone.utc` here as the time provided by Kia seems to be Central European Time.

PR #52, opened 2021-08-02
Correcting TimeZone

cdnninja:


fuatakgun:
thanks for quick action but this should not be needed if I correctly map field names between EU and CA.
in EU response, JSON includes `vehicleStatus.time` attribute on the other hand in CA, it is `status.lastStatusDate'.
Adding your solution will break EU users, I prefer to fix it from source.

PR #57, opened 2021-08-15
Added Pin Support, lock unlock, odometer and location for CAD

cdnninja:
Added pin support for lock unlock.  Testing still required.

### DIFF
--- a/custom_components/kia_uvo/KiaUvoApiCA.py
+++ b/custom_components/kia_uvo/KiaUvoApiCA.py
@@ -6,6 +6,7 @@ import random
 import requests
 from urllib.parse import parse_qs, urlparse
 import uuid
+import json
 
 from .const import DOMAIN, BRANDS, BRAND_HYUNDAI, BRAND_KIA, DATE_FORMAT, VEHICLE_LOCK_ACTION
 from .KiaUvoApiImpl import KiaUvoApiImpl
@@ -21,9 +22,10 @@ class KiaUvoApiCA(KiaUvoApiImpl):
         password: str,
         region: int,
         brand: int,
+        pin: int,
         use_email_with_geocode_api: bool = False,
     ):
-        super().__init__(username, password, region, brand, use_email_with_geocode_api)
+        super().__init__(username, password, region, brand, pin, use_email_with_geocode_api)
 
         if BRANDS[brand] == BRAND_KIA:
             self.BASE_URL: str = "www.myuvo.ca"
@@ -51,6 +53,7 @@ class KiaUvoApiCA(KiaUvoApiImpl):
     def login(self) -> Token:
         username = self.username
         password = self.password
+        pin = self.pin
 
         ### Sign In with Email and Password and Get Authorization Code ###
 
@@ -109,6 +112,19 @@ class KiaUvoApiCA(KiaUvoApiImpl):
         vehicle_status = {}
         vehicle_status["vehicleStatus"] = response
         vehicle_status["vehicleStatus"]["time"] = response["lastStatusDate"]
+        service_status = {}
+        service_status = self._get_service_status(token)
+        vehicle_status["odometer"] = {}
+        vehicle_status["odometer"]["unit"]= service_status["serviceStatus"]["currentOdometerUnit"]
+        vehicle_status["odometer"]["value"]= service_status["serviceStatus"]["currentOdometer"]
+        
+        vehicle_location = {}
+        vehicle_location = self._get_vehicle_location(token)
+        vehicle_status["vehicleLocation"] = {}
+        vehicle_status["vehicleLocation"] = vehicle_location["vehicleLocation"]
+
+        _LOGGER.debug(f"{DOMAIN} - Full Vehicle Status: {vehicle_status}")
+
         return vehicle_status
 
     def update_vehicle_status(self, token: Token):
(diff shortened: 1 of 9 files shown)


cdnninja:
@fuatakgun.  That is enough for one night.  So here is where this is at.  

- verify_pin method fails with a status code 200 back from the server.  I actually think I am messing up the token by passing the token within the method.   The json fails to parse the response. Could you take a look? 
- if I comment out the need for the verify_pin I the following.  Even with hard coded pin number. I suspect because I didn't validate first and pass that with the header to lock. 
- `2021-08-14 20:21:04 DEBUG (SyncWorker_2) [custom_components.kia_uvo.KiaUvoApiCA] kia_uvo - Received lock_action response {'error': {'errorCode': '7310', 'errorDesc': 'PIN incorrect.  Multiple incorrect attempts will require PIN reset'}, 'responseHeader': {'responseCode': 1, 'responseDesc': 'Failure'}}`  
- I added pin support to EU as well.  I wouldn't mind testing to make sure it doesn't break anything.  Not sure EU has value in PIN though. I didn't implement the usage just pass it to the class.

cdnninja:
Well just kidding.  Removed the integration and set it up again and this all works great.   Not sure what changed.  Only thing that needs work is it doesn't refresh and validate it worked until next polling cycle. 

Could you test on the EU side I didn't break anything?  Then the question is how to make it so people can upgrade without uninstall re-install.

cdnninja:
I attempted to get odometer working in CAD.   No errors occur but it doesn't display so I suspect I am not placing the value in the right spot.  Are you able to take a look?

cdnninja:
@fuatakgun wanted to check in to see when you may be able to review?  Hoping to do more commits but wanted to get this big one in first.

fuatakgun:
Thanks for great work, I am on vacation for next 2 weeks and I want to be sure that this is not breaking at least EU/Kia (my car) use cases before merging.

I have one general advice, can you squash all these changes into one commit? so in case anything goes wrong, rolling back can be easier and faster rather than finding correct spot.

cdnninja:
Updated to a single check in.

custom_components/kia_uvo/KiaUvoApiCA.py
```diff
@@ -21,9 +22,10 @@ def __init__(
         password: str,
         region: int,
         brand: int,
+        pin: int,
```
fuatakgun, line 25: is this really integer? can it start with `zero`?

custom_components/kia_uvo/KiaUvoApiCA.py
```diff
@@ -109,6 +112,19 @@ def get_cached_vehicle_status(self, token: Token):
         vehicle_status = {}
         vehicle_status["vehicleStatus"] = response
         vehicle_status["vehicleStatus"]["time"] = response["lastStatusDate"]
+        service_status = {}
+        service_status = self._get_service_status(token)
```
fuatakgun, line 116: as of now, I am not following `_` (underscore naming pattern) ,you can use it without it.

custom_components/kia_uvo/KiaUvoApiCA.py
```diff
@@ -51,6 +53,7 @@ def __init__(
     def login(self) -> Token:
         username = self.username
         password = self.password
+        pin = self.pin
```
fuatakgun, line 56: you need to store this inside `token.set` as a last variable, which can be used for creating Token instance from json

custom_components/kia_uvo/KiaUvoApiEU.py
```diff
@@ -28,8 +28,9 @@ def __init__(
         region: int,
         brand: int,
         use_email_with_geocode_api: bool = False,
+        pin: int,
```
fuatakgun, line 31: no need to introduce this as of now, just set a default value

custom_components/kia_uvo/KiaUvoApiEU.py
```diff
@@ -28,8 +28,9 @@ def __init__(
...
     ):
-        super().__init__(username, password, region, brand, use_email_with_geocode_api)
+        super().__init__(username, password, region, brand, use_email_with_geocode_api, pin)
```
fuatakgun, line 33: keep the correct constructor signature, pin should come after `password` as it is in CA implementation

custom_components/kia_uvo/KiaUvoApiCA.py
```diff
@@ -21,9 +22,10 @@ def __init__(
         password: str,
         region: int,
         brand: int,
+        pin: int,
         use_email_with_geocode_api: bool = False,
     ):
-        super().__init__(username, password, region, brand, use_email_with_geocode_api)
+        super().__init__(username, password, region, brand, pin, use_email_with_geocode_api)
```
fuatakgun, line 28: can we move `pin` after `password`

custom_components/kia_uvo/Vehicle.py
```diff
@@ -1,203 +1,195 @@
...
-from .const import *
+from .const import DOMAIN, BRANDS, BRAND_HYUNDAI, BRAND_KIA, DATE_FORMAT, VEHICLE_LOCK_ACTION
 from .KiaUvoApiImpl import KiaUvoApiImpl
 from .Token import Token
 
 _LOGGER = logging.getLogger(__name__)
 
-class Vehicle(object):
+
+class KiaUvoApiCA(KiaUvoApiImpl):
```
fuatakgun, line 18: I am lost here, why did you put this class into Vehicle.py ?

fuatakgun:
hey, one question, if you want, I can create a new revision on this and we can review together. Is it fine for you?

cdnninja:
Yes for sure.  I’m no developer so just hacking my way through it! 

One note I am hitting some bugs on refresh with my current code.  Looks like I’m maxing the API because of the location and pin check.

custom_components/kia_uvo/Vehicle.py
cdnninja: Looks like a copy and paste mistake.  This was not what was running locally on my side.  Corrected now.

custom_components/kia_uvo/KiaUvoApiEU.py
cdnninja: Done, I know you are making changes anyway but good practice for me.

cdnninja:
I fixed a few items and also gave you edit access.  So feel free to make commits against this pull request.  Or start a new one either way!

PR #65, opened 2021-09-18
Added remote start to CAD and Entity Filters

cdnninja:
Added remote start

Filters binary entities to only ones available on that vehicle.

Remote start is not fully tested.

### DIFF
--- a/custom_components/kia_uvo/KiaUvoApiCA.py
+++ b/custom_components/kia_uvo/KiaUvoApiCA.py
@@ -124,9 +124,9 @@ class KiaUvoApiCA(KiaUvoApiImpl):
         vehicle_status["odometer"] = {}
         vehicle_status["odometer"]["unit"]= response["currentOdometerUnit"]
         vehicle_status["odometer"]["value"]= response["currentOdometer"]
-
         vehicle_status["vehicleLocation"] = self.get_location(token)
 
+            
         return vehicle_status
 
     def get_location(self, token: Token):
@@ -138,8 +138,9 @@ class KiaUvoApiCA(KiaUvoApiImpl):
 
         response = requests.post(url, headers=headers, data=json.dumps({"pin": self.pin}))
         response = response.json()
-
+        
         _LOGGER.debug(f"{DOMAIN} - Get Vehicle Location {response}")
+
         return response["result"]
 
     def get_pin_token(self, token: Token):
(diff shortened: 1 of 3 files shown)


custom_components/kia_uvo/KiaUvoApiCA.py
```diff
@@ -179,13 +179,33 @@ def lock_action(self, token: Token, action):
...
+        url = self.API_URL + "rmtstp"
+        headers = self.API_HEADERS
+        headers["accessToken"] = token.access_token
+        headers["vehicleId"] = token.vehicle_id
+        headers["pAuth"] = self.get_pin_token(token)
+
+        response = requests.post(url, headers=headers, data=json.dumps({"pin": self.pin}))
+        response = response.json()
+
+        _LOGGER.debug(f"{DOMAIN} - Received start_car response {response}")
```
fuatakgun, line 207: typo: stop_car

custom_components/kia_uvo/binary_sensor.py
```diff
@@ -37,17 +38,19 @@ async def async_setup_entry(hass, config_entry, async_add_entities):
...
         BINARY_INSTRUMENTS.append(("charging", "Charging", "vehicleStatus.evStatus.batteryCharge", None, None, DEVICE_CLASS_BATTERY_CHARGING))
         BINARY_INSTRUMENTS.append(("pluggedIn", "Plugged In", "vehicleStatus.evStatus.batteryPlugin", None, None, DEVICE_CLASS_PLUG))
     if vehicle.engine_type is VEHICLE_ENGINE_TYPE.PHEV or vehicle.engine_type is VEHICLE_ENGINE_TYPE.IC:
         BINARY_INSTRUMENTS.append(("lowFuelLight", "Low Fuel Light", "vehicleStatus.lowFuelLight", "mdi:gas-station-off", "mdi:gas-station", None))
 
-    binary_sensors = [
-        InstrumentSensor(hass, config_entry, vehicle, id, description, key, on_icon, off_icon, device_class)
-        for id, description, key, on_icon, off_icon, device_class in BINARY_INSTRUMENTS
-    ]
+    binary_sensors = []
```
fuatakgun, line 50: what about like this (please pay attention to spaces/indentation)
```
for id, description, key, on_icon, off_icon, device_class in BINARY_INSTRUMENTS:
    if vehicle.get_child_value(key) != None:
        binary_sensors.append(InstrumentSensor(hass, config_entry, vehicle, id, description, key, on_icon, off_icon, device_class))
```

```

File: custom_components/kia_uvo/KiaUvoApiCA.py [hunk 3]
```diff
@@ -124,9 +124,11 @@ def get_cached_vehicle_status(self, token: Token):
         vehicle_status["odometer"] = {}
         vehicle_status["odometer"]["unit"]= response["currentOdometerUnit"]
         vehicle_status["odometer"]["value"]= response["currentOdometer"]
-
-        vehicle_status["vehicleLocation"] = self.get_location(token)
-
+        try:
```
fuatakgun, line 127: let's postpone this, as we do not have any try/catch mechanism in overall integration, we can discuss what could be the best.
- having try catch in final functions: from development perspective, this is not good but easy to handle, one place to catch the errors
- having try catch in caller functions: from development perspective, this looks fine, but we might need to catch in multiple places depending on functions

fuatakgun (changes requested):

custom_components/kia_uvo/KiaUvoApiCA.py
cdnninja: Fixed this and the other one.

custom_components/kia_uvo/KiaUvoApiCA.py
cdnninja: Done.  This one is important to me as I have to comment this line out to test.  The line works when not called too often though.

custom_components/kia_uvo/binary_sensor.py
cdnninja: Done.  Testing now, I also did the same for "sensors" as items like range for fuel don't exist on mine.

cdnninja:
@fuatakgun  ready for another round of review.

fuatakgun (approved):
please share your test results, ready to merge when you are ready

cdnninja:
Worked no issue.   Only item I am finding is "last updated" doesn't show. However it never worked properly before.   I propose merging.

fuatakgun:
> Worked no issue. Only item I am finding is "last updated" doesn't show. However it never worked properly before. I propose merging.

Thanks for your support and patience, much appreciated

PR #66, opened 2021-09-18
Bug fix for updated time filter

cdnninja:
Corrected time zone and exempted it from filters.   Please validate this doesn't break europe, I am assuming it is okay.

### DIFF
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -42,7 +42,7 @@ async def async_setup_entry(hass, config_entry, async_add_entities):
     sensors = []
 
     for id, description, key, unit, icon, device_class in INSTRUMENTS:
-        if vehicle.get_child_value(key) != None:
+        if vehicle.get_child_value(key) != None or key == "last_updated":
             sensors.append(InstrumentSensor(hass, config_entry, vehicle, id, description, key, unit, icon, device_class))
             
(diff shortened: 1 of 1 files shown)


custom_components/kia_uvo/Vehicle.py
```diff
@@ -173,7 +173,7 @@ def set_last_updated(self):
             hour=int(m.group(4)),
             minute=int(m.group(5)),
             second=int(m.group(6)),
-            tzinfo=local_timezone,
+            tzinfo=dt_util.get_time_zone("UTC"),
```
fuatakgun, line 176: This is not valid for EU because response from kia servers return time in local time, not in UTC. Can you confirm response for CA in UTC? in any case, let's separate these 2 changes as sensors.py is a quick bug fix, which we introduced recently.

fuatakgun (changes requested):
Please keep sensors.py as a quick bug fix

custom_components/kia_uvo/Vehicle.py
cdnninja: Yes CA is in UTC.  Before this change my updates were reporting as in the future.

fuatakgun: is it fine for you if you can separate this into another PR?

fuatakgun (approved):

PR #67, opened 2021-09-18
Corrected time zone for CAD last updated

cdnninja:


cdnninja:
Assuming europe is always Europe/Berlin we will need to add logic to handle this.

fuatakgun:
fixed with latest commit, please test and share results

cdnninja:
Confirmed it works.   However set temp reports 16.5 - no way my wife has it set to that. I will dig in.

PR #70, opened 2021-09-20
Error handling for location API

cdnninja:
Would like to have you review this approach first.


custom_components/kia_uvo/Vehicle.py
```diff
@@ -98,7 +98,7 @@ def set_geocoded_location(self, old_vehicle_location):
         new_lat = self.get_child_value("vehicleLocation.coord.lat")
         new_lon = self.get_child_value("vehicleLocation.coord.lon")
 
-        if (old_lat != new_lat or old_lon != new_lon) or old_geocode is None:
+        if ((old_lat != new_lat or old_lon != new_lon) or old_geocode is None) and (old_lon != None and old_lat != None):
```
fuatakgun, line 101: Why do we need this additional condition? Enabling geolocation could be done using options and just before enabling it, it could be none.

custom_components/kia_uvo/KiaUvoApiCA.py
```diff
@@ -124,7 +124,12 @@ def get_cached_vehicle_status(self, token: Token):
         vehicle_status["odometer"] = {}
         vehicle_status["odometer"]["unit"]= response["currentOdometerUnit"]
         vehicle_status["odometer"]["value"]= response["currentOdometer"]
-        vehicle_status["vehicleLocation"] = self.get_location(token)
+        
+        vehicle_status["vehicleLocation"] = {}
+        try:
```
fuatakgun, line 129: i personally would have try/catch block inside get_location function and return empty dictionary `{}` for any error. So all callers would benefit it. 

What happens if we don't have lat and lon items in vehicle_location? Any further error? Maybe we can create these items on empty response.

fuatakgun (changes requested):
Requested changes

custom_components/kia_uvo/Vehicle.py
cdnninja: Typo - this should be checking for new_lon, was crashing before because new was empty.

custom_components/kia_uvo/KiaUvoApiCA.py
cdnninja: I have moved the try catch. I don't want to return invalid location as the last known location may still be correct.  Thoughts?

cdnninja:
This try catch location didn't work today as the API did return something, just an error.  I will try a different approach.

cdnninja:
Updated to throw an exception if data is missing.   Works currently but my API for location is working no issue.  Considering number of reboots to test this I suspect my API will max later today to test the failure.

custom_components/kia_uvo/KiaUvoApiCA.py
```diff
@@ -123,25 +123,31 @@ def get_cached_vehicle_status(self, token: Token):
...
-        _LOGGER.debug(f"{DOMAIN} - Get Vehicle Location {response}")
-
-        return response["result"]
+        try:
+            headers["pAuth"] = self.get_pin_token(token)
+
+            response = requests.post(url, headers=headers, data=json.dumps({"pin": self.pin}))
+            response = response.json()
+            _LOGGER.debug(f"{DOMAIN} - Get Vehicle Location {response}")
+            if response["result"]["coord"]["lat"] == None:
```
fuatakgun, line 142: if the key does not exists, it will try another error. key means `lat` or `coord`. i suggest you to check result status and status code. there is always S (succees) and F (failure) exists in the response. until we find a better solution for this, we can disable this call from parent function (you can keep it locally for testing purposes)

if you need a hand, happy to jump in but I do not have any account in CA :) if you can share your car for couple of hours, we can work together. or else, you can dump response of API.

fuatakgun (changes requested):

custom_components/kia_uvo/KiaUvoApiCA.py
cdnninja: I can check but I don't think you can share cars in Canada.  

Are you thinking like this? 

`            if response["responseHeader"]["responseCode"] != 0:
                raise Exception('No Location Located')`

fuatakgun: Yes similar to this.

cdnninja: Done.  This has come along when with the latest commit.   Had to handle a few spots to try make this not break other parts.  So far been pretty stable for me.

custom_components/kia_uvo/KiaUvoApiCA.py
```diff
@@ -188,7 +200,7 @@ def start_climate(self, token: Token):
 
         response = requests.post(url, headers=headers, data=json.dumps({"setting": {"airCtrl": 0, "defrost": "false", "heating1": 0, "igniOnDuration": 3, "ims": 0}, "pin": self.pin}))
         response = response.json()
-
+        time.sleep(3)
```
fuatakgun, line 203: Why do you need this?

fuatakgun (commented):
So, as a quick solution, i am still inclined to disable these functions (location and odometer). We should build a solution like this;
- check previous last update time with latest
- if there is a change, check location and odometer
- if not, use existing values

Otherwise, we will always hit the limit and it is dangerous.

cdnninja:
Would you be open to a 15-30 minute call at some point?  It may be easier to walk you through what I am seeing from the Canadian API.

fuatakgun:
Hey, it is late in EU now, but what about tomorrow? Please drop me an email over fuatakgun@gmail.com and we can have a google meet session.

custom_components/kia_uvo/KiaUvoApiCA.py
cdnninja: The API takes a good while to actually function.  This seems to help so we don't move on and call the service again.   I may be able to remove these with the adjustments I just made though.

cdnninja:
Changed approach on this.  The only APIs I hit limits on are items I pass the PIN for, maintenance isn't one of them.   So I am using odometer as something to compare with.  If this hasn't increased I don't call location.   Now I am not sure holding this data in the API class is the best spot for it.  Still open a call to discuss.  I am no pro just messing around. :)

cdnninja:
I will monitor this new change tomorrow morning as my location API is maxed out.    Now everything else works great without errors which is how it should be.  Ideally tomorrow location works fine which means we only call this occasionally and can also handle errors if it ever is maxed out.

cdnninja:
Closing this pull to create a cleaner final one.

Issue #88, opened 2021-10-17
KIA USA Support


cdnninja:
To continue the conversation on #80 specific for Kia. @mbbush.

mbbush:
I put the mock responses file at 
https://github.com/mbbush/kia_uvo/pull/1

mbbush:
I can see a request to `https://owners.kia.com/apps/services/owners/getvehicleinfo.html/vehicleFeature/1/vehicleStatus/1/enrollment/1/maintenance/1/dtc/1` which uses that cookie, and a header named `vinkey`, which is a uuid, although I haven't figured out where it comes from yet. This request returns json containing lots of useful data about the car.

Still TODO:
* Figure out how to refresh the data
* make some sort of more organized collection of requests/responses and share them in a useful format. I use insomnia.rest at work for sharing "here's an api you can call manually", so I'll probably create an insomnia collection. It's the same idea as postman, but with some cool extra features. 

How similar or dissimilar does this sound to the way it works in CA or EU? I haven't read the existing source code yet; I should probably do that...

mbbush:
ok, so this api is funky.

To login, you have to make a POST request, with a JSON body, but set the content type to `application/x-www-form-urlencoded`. All other combinations of sending the data I've tried (form data sent with the content type for form data, json with json content type), you don't get a response body, but you still get the session cookie. The one other thing that worked was sending form data with a single key consisting of the url-encoded json body, and no value. Yuck. 

Anyways, this works, and returns a body and a session cookie. Still working on the subsequent requests.
```
import requests
import json

payload = { 
  "userId": "user",
  "password": "PASSWORD",
  "userType": "1",
  "vin": "",
  "action": "authenticateUser"
  }
 
 r = requests.post('https://owners.kia.com/apps/services/owners/apiGateway', data=json.dumps(payload), headers={'Content-Type': 'application/x-www-form-urlencoded'})
 ```

cdnninja:
It may be worth comparing to what the android app does.  It looks like this is front ended by aws amazon gateway.  If we can crack the key like the Hyundai usa integration I bet it will be easier in the long haul. 

It seems like two apis exist.  The website and the app.

mbbush:
I'll get started on spying on the app, then. It's been a long time since I've dealt with intercepting SSL traffic, so I'm not sure how long it'll take me. 

I did see a couple properties in the manifest file that looked like they could be flipped to allow insecure requests/easy debugging, so maybe I'll try that first.

[...]

PR #90, opened 2021-10-18
Kia USA Login and Vehicle Retrieval

mbbush:
#88


cdnninja (commented):
@fuatakgun  I am okay with this if you are.  Not sure if a better way exists to handle the token situation?  At this point it is only into the USA branch which once semi functional will merge with the regular code base.

PR #91, opened 2021-10-20
fix: Misc changes to succesfully parse the Hyundai USA payload into sensors

bvlaicu:
Fixes https://github.com/fuatakgun/kia_uvo/issues/80

The following sensors are created successfully (that's all except `temperatureSetpoint` which for the USA API is not a number)

image

### DIFF
--- a/custom_components/kia_uvo/HyundaiBlueLinkAPIUSA.py
+++ b/custom_components/kia_uvo/HyundaiBlueLinkAPIUSA.py
@@ -79,12 +79,13 @@ class HyundaiBlueLinkAPIUSA(KiaUvoApiImpl):
         response = requests.get(url, headers=headers)
         _LOGGER.debug(f"{DOMAIN} - Get Vehicles Response {response.text}")
         response = response.json()
-        vehicle_name = response["enrolledVehicleDetails"][0]["vehicleDetails"]["nickName"]
-        vehicle_id = response["enrolledVehicleDetails"][0]["vehicleDetails"]["vin"]
-        vehicle_model = response["enrolledVehicleDetails"][0]["vehicleDetails"]["modelCode"]
-        vehicle_registration_date = response["enrolledVehicleDetails"][0]["vehicleDetails"]["enrollmentDate"]
+        vehicle_details = response["enrolledVehicleDetails"][0]["vehicleDetails"]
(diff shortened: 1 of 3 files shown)


cdnninja:
Great work!  On the comment of temp in EU and Canada it reports the lowest value unless the car is running.  Looks like in the USA that would be LOW, I suspect you would get a real value with the car running with a remote start.   On the to do is to filter this data out when the car isn't running.   Luckily looks like USA reports real data vs the hex that comes from EU and CA. 

Could you check if my thoughts could be true for USA temp?

custom_components/kia_uvo/HyundaiBlueLinkAPIUSA.py
```diff
@@ -106,13 +107,20 @@ def get_cached_vehicle_status(self, token: Token):
         url = self.API_URL + "rcs/rvs/vehicleStatus"
         headers = self.API_HEADERS
         headers["accessToken"] = token.access_token
+        headers["vin"] = token.vehicle_id
+        headers["refresh"] = "false"
```
cdnninja, line 111: Could we try move this to the api_headers?  This will let you not need to repeat it once the remaining methods to be built out.  It will depend if it breaks login or not.

custom_components/kia_uvo/sensor.py
```diff
@@ -48,7 +51,8 @@ async def async_setup_entry(hass, config_entry, async_add_entities):
     INSTRUMENTS.append(("geocodedLocation", "Geocoded Location", "vehicleLocation.geocodedLocation.display_name", None, "mdi:map", None))
     INSTRUMENTS.append(("carBattery", "Car Battery", "vehicleStatus.battery.batSoc", PERCENTAGE, "mdi:car-battery", DEVICE_CLASS_BATTERY))
 
-    INSTRUMENTS.append(("temperatureSetpoint", "Set Temperature", "vehicleStatus.airTemp.value", TEMP_CELSIUS, None, DEVICE_CLASS_TEMPERATURE))
+    if config_entry.data.get(CONF_REGION, DEFAULT_REGION) is not REGION_USA:
```
cdnninja, line 54: Are you able to confirm if this reports correctly when the car is running?  If so it may be the same as EU and CA. Would rather fix it for all regions if that is the case.

PR #97, opened 2021-10-31
feat: Engine start and stop for Hyundai USA

bvlaicu:
Related to https://github.com/fuatakgun/kia_uvo/issues/80
Includes a global change to add the vehicle registration id to the Token class.


cdnninja:
Thanks for this!  Great to see some USA progress. With the current services.yaml only supporting celsius temp entries are you just using celsius commands or are you assuming people with override the GUI?  I am not sure a good solution exists here for the HA GUI.  I will take a look to see what other integrations are doing.

PR #101, opened 2021-11-04
feat: Added EV to CA

cdnninja:
Corrected EV remote start/stop for CA and added charge start stop.


cdnninja:
@Kevr0n228 are you able to do a final test against this version?  Should work for charging stop start and climate stop start.

Kevr0n228:
Stop_climate works for me, have not had a chance to test charging though.

PR #105, opened 2021-11-08
fix: Corrected Kia USA Token model name and fix engine type for Niro EV

dahlb:


cdnninja:
Could you split this into two pulls? The API changes are an easy merge but the logic on ev needs to be tested in other regions. 

We have fought with that logic before between Canada and EU.

dahlb:
closing for split PRs

PR #112, opened 2021-11-09
feat: populate location for USA Kia for use in device_tracker

dahlb:


dahlb:
please merge #110 first, I think I prevented conflicts but if I didn't I'd rather fix on the smaller branch

PR #117, opened 2021-11-09
fix: Corrected remote start types

cdnninja:


dahlb:
seems it would be cleaner to move this logic into the CA implementation of start_climate to call it's ev method internally if engine is ev

cdnninja:
If so I need to pass engine type into the API as the API isn't aware of engine type. I actually did this in a previous version but removed that type of logic as I have to decode and store engine type in both vehicle and api.   Or pass it into the start call.  Open to suggestions.

dahlb:
I haven't used those functions yet but based on initially look, with this and some hacks to track current version of vehicle maybe these or at least some of the api fuctions would benefit from adding vehicle object to signatures

cdnninja:
I am merging this to get it fixed as a temp fix.    @fuatakgun  is going to put some time into how to refactor some of this.  We are hitting issues with different APIs needing different things.

PR #119, opened 2021-11-10
fix: Clean up spacing on CA API

cdnninja:


cdnninja:
Retested remote start stop as it is a key item that changed indenting.  Still works.

PR #123, opened 2021-11-12
feat: USA Kia reauth, lock, unlock, start and logging responses

dahlb:
I added the fix for stale sessions and catch invalid responses from the server for logging.  I put this as a second PR but with  https://github.com/fuatakgun/kia_uvo/pull/122 included as it builds on those functions and drys them up, if you merge this I'll close 122 or if 122 is merged it will reduce the diff on this PR.

### DIFF
--- a/custom_components/kia_uvo/KiaUvoAPIUSA.py
+++ b/custom_components/kia_uvo/KiaUvoAPIUSA.py
@@ -1,17 +1,13 @@
-import base64
 import logging
-import sys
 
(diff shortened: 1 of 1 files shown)


dahlb:
if you rerun the lint it should pass, I fixed the commit message it's just a stale result.

dahlb:
easiest way to reproduce the stale session issue is to restart HA, login on your phone, try to do anything with kia in HA within the next hour ...

PR #126, opened 2021-11-13
feat: Add Code Lint Checking

cdnninja:
Adds lint checking much like HACS uses.  It also cuts a PR against the source branch of another PR when it fails.  It will be on PR per failed check, for example one for the black fixes one for pyupgrade.  Examples can be viewed here: 

https://github.com/cdnninja/kia_uvo/pull/3
https://github.com/cdnninja/kia_uvo/pull/9
https://github.com/cdnninja/kia_uvo/pull/10

If people are on board with going Black I will first test a compliant version of the repo before we commit this.    I don't see it causing issues but you never know!  Feedback appreciated since this could be annoying,


cdnninja:
@dahlb  Interested in your thoughts as well.

dahlb:
as a new comer to this project and python, that hadn't programmed for a year until this contrib.  

I saw a comment about somebody concerned enough about empty lines to ask for a change in code so black seems a great way to keep things consistent without wasting anybody's time.

  I generally like lints, assuming they work; I've had some lint failures and I think they are obviously the only reason I managed to follow preferred practices with comment messages here.  I'm just used to being able to restart pipelines when they fail, but I think I only had an issue because of correcting things using force pushes.

cdnninja:
That is fair! I am new to this as well.  I jumped in for the Canadian side.  My thoughts are that a decent lint checker helped me contribute to other repos in the past so may be great for this one.  If @fuatakgun  is good with using a black checker I will make this happen.

fuatakgun:
Agree with black, it is also required by home assistant team to be a core integration

cdnninja:
I have the action bot working against the lint-checking branch.  I will get the validations passing on master slowly before I merge the action to the main branch.  This way it doesn't mess with other Pulls.   You will see PRs being created but those auto created are all against the lint branch not master.

custom_components/kia_uvo/KiaUvoAPIUSA.py
```diff
@@ -98,25 +123,29 @@ def api_headers(self) -> dict:
...
-        headers['deviceid'] = self.device_id
+        headers["date"] = date
+        headers["deviceid"] = self.device_id
         return headers
 
     def authed_api_headers(self, token: Token):
         headers = self.api_headers()
-        headers['sid'] = token.access_token
-        headers['vinkey'] = token.vehicle_regid
+        headers["sid"] = token.access_token
```
fuatakgun, line 132: very happy to see that it is converting single quotes into double one :)

fuatakgun (commented):
thanks for this improvement, if we continue with this speed, it is almost impossible to merge this :) let's agree not to merge anything until you can merge this

fuatakgun (approved):

PR #138, opened 2021-11-15
fix: improve readable with line length limits

dahlb:


dahlb:
@cdnninja I thought the new black changes were handled in a pre commit hook? is there something I need to do manually to get the new stuff working?

cdnninja:
Yes it should.   Looks like it can't checkout your fork repo.  I will take a look later and see why.   My pulls are typically from within this project and not from a fork.  I am sure this is something we can fix though.  It won't stop merges while I sort it out though just means it won't enforce the pull request lint.

cdnninja:
This looks to be the same issue: https://github.com/actions/checkout/issues/143

cdnninja:
I fixed the issue that makes it not run on forks.   The PR won't work though it will just show it failed the tests.   Reason being is we don't have access to create a branch in your fork to generate the PR from.  I will also see if I can swap the PR to just a comment in the existing PR.   I would need to find an action for that.   I will read on other options. 

Feel free to use a branch in this repo if you need the PR to be generated.

dahlb:
@cdnninja can these automated fixes be changed to a post-commit hook on master? If I understand correctly at least the black stuff is about automating fixes for styling.  Having to make branches on the main repo for pr seems a high bar, I probably would've just forked and left if that was in place when I first started.

cdnninja:
Agreed. Need to find a way to make it work for a remote fork.   The goal of this is to make it easier, I find it helpful since my formatting may not align. 

The PR was an addition to the code HACS uses themselves to auto resolve.  Looks like most people leave it to the dev to comply and don't help along the way. 

My only concern with post commit is we have no way to over rule if need be.  I'll keep working on this to find a way.  I noticed zwave js has a bot that work on remote forks. 

Right now it doesn't stop anyone and goes all green on a valid commit so while I figure out way forward I think it's okay for a few days.

cdnninja:
Other thing to note the PR is meant to be in your fork to make life easier. This way you see the changes hit merge and it's done. It isn't a requirement.  Without it black just says it failed but not why.

dahlb:
@cdnninja sorry I don't understand your last comment, are you pointing out I could skip the pr and just push to master and black would just mark the master branch as failed?  I prefer to put things in PRs as it makes emails for those watching the repo vs having to track file changes manually through direct master pushes

dahlb:
and I get the concern that black can't be overruled in a post-hook but it seems a widely used tool that shouldn't need monitoring?

cdnninja:
No, what I am saying is the system proposes the changes to your commit in a separate PR.  Now this works within this repo but isn't working for remote.  The idea was you just approve the PR against your commit and it auto resolves your lint issues.   Now the second part isn't working on a fork, only validation works not correction.  

Are you open to a quick google meet?  Happy to chat through this a bit.   If so shoot me an email at jaydenaphillips@gmail.com.

dahlb:
ic, that makes sense, I still think the post commit should be safe with a tool like black but to keep the same workflow maybe it could always create the extra pr in this repo even if the extra pr destination is the remote?  

sorry I can't do a meet I'm on extended parental leave and have been working on this while my 21 month old naps and a little before bed, but they wake up if I talk, so I can only type :)

cdnninja:
Yes that may work.  I am pretty new with github actions so have modeled after other large repos as a starting point, which is why it is setup as a pre-commit currently.  I hear you the concern and agree - the goal is to make this easier.   I will open an issue for us to continue this conversation.  I am currently playing with options so you will see some random PRs against your branch.  We will see what I can make work.

PR #139, opened 2021-11-15
fix: usa low/high temp values

dahlb:


fuatakgun:
added you @dahlb as collaborator to repository, you can push urgent/blocker fixes. for the rest, let's follow review process.

dahlb:
@cdnninja I don't understand why this pr is failing and usa_readable is passing. I ran black manually and it had not changes to the branch.

cdnninja:
Re runs don't use the new config until a new commit is done. So for this one off either we can ignore the failure or tweak something and set it back to get green. 

It's an odd GitHub thing.

dahlb:
@tchoward can you verify this fix, you're the only one that's reported experiencing this bug

cdnninja:
I am thinking we merge this if we don't hear back as it only impacts one region and sounds like you have tested that it atleast doesn't seem to break your instance.

tchoward:
Sorry for not responding; been a week.  In any case, I am not seeing this issue any more.

PR #141, opened 2021-11-16
feat: use ha distance unit

dahlb:


cdnninja:
I ran into a chat yesterday on talking about the UK using both celsius and miles. This was in the home assistant community thread.    I know I was originally the one thinking we should use the HA setting.     I wonder if that was the reason to only override distance?

fuatakgun:
Hey, you can default to HA settings but do not remove existing functionality

dahlb:
closing, new solution in https://github.com/fuatakgun/kia_uvo/pull/149 seems an odd use case but what do I know I'm stuck using imperial units ...

PR #144, opened 2021-11-16
fix: battery zero incorrect

dahlb:


custom_components/kia_uvo/Vehicle.py
```diff
@@ -64,7 +65,16 @@ async def update(self):
                     self.set_geocoded_location, current_vehicle_location
                 )
 
-            async_dispatcher_send(self.hass, self.topic_update)
+            if (
+                self.get_child_value("vehicleStatus.engine") == False
+                and previous_vehicle_status.engine == False
```
fuatakgun, line 70: Does this line really work? Have you tried this?

custom_components/kia_uvo/Vehicle.py
```diff
@@ -64,7 +65,16 @@ async def update(self):
...
+                and self.get_child_value("vehicleStatus.evStatus.batteryStatus") == 0
+                and previous_vehicle_status["evStatus"]["batteryStatus"] != 0
+            ):
+                await self.force_update()
```
fuatakgun, line 74: Let's fire this as a separate request rather than waiting results

dahlb: @fuatakgun I can switch this ti a second request, but the reason I did it with waiting is so that scripts in HA that call the update function are able to tell when it actually finishes.  If this is changed to async the scripts will call update thinking the values are updated but they won't be until the force update call finishes.

custom_components/kia_uvo/Vehicle.py
dahlb: @fuatakgun nope that syntax is my old ruby showing, I've updated it to correct python syntax and verified it causes no regressions anymore locally, it will take a day or two in order to verify fully

custom_components/kia_uvo/Vehicle.py
fuatakgun: ok, I see your point

custom_components/kia_uvo/Vehicle.py
```diff
@@ -53,6 +53,7 @@ def __init__(
 
     async def update(self):
         try:
+            previous_vehicle_status = self.get_child_value("vehicleStatus")
```
fuatakgun, line 56: can we follow same naming below (current_vehicle_location). I am a bit obsessed with naming :) 
image

dahlb: @fuatakgun I moved the definition of the currentLocation below the impl update call, I think it isn't so much a naming consistency issue as a bug. I couldn't think of a reason to update the location using the previous location during an update instead of the current.

fuatakgun: There is no reason to call external endpoint if location coordinates have not changed, right?

fuatakgun: My comment is about naming; if we call current location as current, i was saying we should call current vehicle data.

dahlb: ic, it didn't make sense. until I saw the parameter name is old_vehicle_location, previous_vehicle_location seems more inline with old_vehicle_location then current, and make more sense without having to understand it's use.

PR #150, opened 2021-11-17
feat: track climate finished

dahlb:
will resolve https://github.com/fuatakgun/kia_uvo/issues/142

### DIFF
--- a/custom_components/kia_uvo/Vehicle.py
+++ b/custom_components/kia_uvo/Vehicle.py
@@ -168,9 +168,15 @@ class Vehicle:
                 climate,
                 heating,
             )
+        # call update to see climate started
         self.force_update_try_count = 0
         self.force_update_try_caller = async_call_later(
             self.hass, START_FORCE_UPDATE_AFTER_COMMAND, self.force_update_loop
+        )
+
+        # call update to see climate finished
+        async_call_later(
+            self.hass, (duration * 60) + 1, self.force_update_loop
         )
 
     async def stop_climate(self):
--- a/custom_components/kia_uvo/Vehicle.py
+++ b/custom_components/kia_uvo/Vehicle.py
@@ -175,7 +175,11 @@ class Vehicle:
         )
 
         # call update to see climate finished
-        async_call_later(self.hass, (duration * 60) + 1, self.force_update_loop)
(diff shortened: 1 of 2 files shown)


cdnninja:
Only thought on this is it may not work for EU since duration is a NA thing.  In the EU climate will run for a preset length of time.  Now I think we are okay on that since the default is set to 5 so it would check in 5 minutes later to see if it finished.  I will sift through code to confirm the theory since I can't test to confirm.  

Great idea for an add by the way!

dahlb:
@cdnninja great point on the duration being ignored, I hadn't thought to check and in the USA it is ignored too. I haven't worked out how long it is yet exactly but seems more then 5 and likely is 15 minutes.  Going to have to rethink this feature a bit once I pin point the duration in USA kia.  I really wanted it for a script to keep the climate on until I turn it off or the car turns on, sometimes I just leave it running for hours unlocked and that would be much better ;)

custom_components/kia_uvo/Vehicle.py
```diff
@@ -168,11 +168,15 @@ async def start_climate(self, set_temp, duration, defrost, climate, heating):
                 climate,
                 heating,
             )
+        # call update to see climate started
         self.force_update_try_count = 0
         self.force_update_try_caller = async_call_later(
             self.hass, START_FORCE_UPDATE_AFTER_COMMAND, self.force_update_loop
         )
 
+        # call update to see climate finished
```
fuatakgun, line 177: I am fine with this implementation, I would just increase the buffer (+1), keep a reference to async_call_later, in case there is a new `start_climate` call, so I can cancel the previous schedule and schedule a new one.

A different solution could be `set a milestone in the future if it is still ON, if yes, call force_update`. not a blocker though.

fuatakgun (approved):

custom_components/kia_uvo/Vehicle.py
dahlb: updated buffer and added reference, waiting to merge until duration issue in USA addressed

dahlb:
just observed 18 minute climate run, US definitely ignores the parameter, will have to run more tests to see how consistent it is

PR #153, opened 2021-11-17
fix: reduce force refreshing when not needed

dahlb:
related to https://github.com/fuatakgun/kia_uvo/issues/151 this works for the USA kia to avoid calling and waiting to finish the force refresh endpoint since the action is known to have completed and the api has already cached that result. @cdnninja this will probably also work for the CA version but I didn't want to change the impl flag since I can't test it.

### DIFF
--- a/custom_components/kia_uvo/KiaUvoAPIUSA.py
+++ b/custom_components/kia_uvo/KiaUvoAPIUSA.py
@@ -12,21 +12,15 @@ import time
 
 from .const import (
     DOMAIN,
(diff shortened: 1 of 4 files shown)


dahlb:
@fuatakgun given the restrictions you mentioned for force refreshes I'd like to get this merged soon so the USA users can stop hitting that endpoint a bunch of unneeded times, I've tested every action using this updated code and the results are always up to date since the USA impl waits until the actions are verified completed

PR #157, opened 2021-11-18
feat: Lint Black Auto Correction for Remote Fork

### LINKED ISSUE #143: Github Actions Lint
To continue discussions around future github actions.  

I wanted to get your thoughts @dahlb  on this demo: https://github.com/peter-evans/slash-command-dispatch/pull/28

It looks like it can get around the github limitation arounds pushing commits.  So if implemented I would think we do the following:

1. Keep the existing checks, they wouldn't "fix" anything just tell you if issues exist.
2. Comment in the PR someone created to let them know /black or others are available to them. 
3. If the run /black it resolves the issue in the commit and checks restart. 

I haven't got this working yet in my repo but it seems to work in the demo.

cdnninja:
A token needs to be created and stored as  secrets.PAT before this can be merged.   Closes #143.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -26,7 +26,6 @@ jobs:
         with:
           ref: ${{github.event.pull_request.head.ref}}
           repository: ${{github.event.pull_request.head.repo.full_name}}
-          token: ${{ secrets.PAT }}
       - name: Set up Python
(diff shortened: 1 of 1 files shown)


cdnninja:
Removed token for now.   Token will let checks re-run a second time to ensure everything is fixed.   Once token is added I will add it to the workflow.

PR #170, opened 2021-11-23
feat: Action status tracking

dahlb:


dahlb:
this supersedes https://github.com/fuatakgun/kia_uvo/pull/153 and removes the sleeps from CA and USA which I've learned are considered bad form in async home assistant patterns because they hold onto a worker for the duration of the sleeping.  @cdnninja when you have a chance can you see if this works for you, I made a bunch of assumptions about the CA status checking being the same as the USA status checking; open to other solutions and feedback just looking to reduce force updates which at least in USA aren't needed when you've waited for the action to complete and looking to remove the sleep calls

cdnninja:
I will give this a test, hopefully tonight.

cdnninja:
Works for lock unlock here.  I haven't tried remote start but it is the same code.

dahlb:
@cdnninja did you notice with the lock/unlock if the HA status of the lock was updated after the action finished? besides just making sure the action code still works the other part is that it no longer uses force update at all after actions complete.  This works in US and I'm guessing it will work as you status endpoints look very similar to the USA ones, but skipping force update is a big change that I want to make sure still updates after the lock/unlock :)

cdnninja:
Yes it looks to update status correctly as well.   I am happy to merge this if you are.

PR #175, opened 2021-11-24
fix: Corrected Version Numbers in HA

### LINKED ISSUE #172: Update Version in manifest.json for release workflow
HASC lists the git version but HA doesn't have matching version, having the HA version being accurate could be very helpful in bug reporting.

https://github.com/jossef/action-semantic-release-info creates a github workflow variable of ${{steps.semantic_release_info.outputs.version}} that we should be able to sed or something similiar into the manifest.json file.

cdnninja:
Resolves #172

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release
 on:
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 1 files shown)


dahlb:
👍

PR #183, opened 2021-11-26
feat: add binary sensor to track the action API status

### LINKED ISSUE #180: Track actions to completion with binary sensor
**Is your feature request related to a problem? Please describe.**
The API fails if a second action is called before the first action is considered completed.

**Describe the solution you'd like**
Providing a binary sensor API Action in Progress to indicate when it is and isn't safe to invoke actions

**Describe alternatives you've considered**
I attempted to re-implement the integration to be synchronous but it seems fundamentally impossible due to the API need to wait with calls to async_call_later

dahlb:
tracks start and finish and which action is in progress

closes #180

this handles the initial tracking of actions, I've verified this works with everything except charging (as my car isn't currently plugged in) but see no reason it would act differently.  

The entity is very timely and is instantly updated with the status as an action starts and when it eventually finishes after the "set and forget" is underway to verify the action has completed in the KIA's async api through tracking the api's xid (action's async transaction id).

 it is only available to implementations that track the action's xid, but won't show for users of the other implementations. 

I also did a small optimization in using a separate topic as this is not related to the update method (the current topic is all about update) and I don't want to wait HA entity updates.

I store the entity's data locally to ensure only the event loop modifies the data referenced by entities, if a worker updates the data while the event loop is reading it the other entities could cause a fault.


dahlb:
this also adds a lock for the action tracking

this is the api response if you invoke a second action before the first is completed by the kia api (you can only see the results if you track them like the CA and US implementations are doing.

got response {"status":{"statusCode":1,"errorType":1,"errorCode":1125,"errorMessage":"Another remote command is being executed. Please wait for it to finish and try again."}}

PR #199, opened 2021-12-04
feat:  Support for multiple accounts

cdnninja:
A draft PR to support multiple accounts.  This isn't working just yet but the PR will show the direction.  To do:

1. Adjust config flow to block if same details are used?   Or ensure it replaces the previous if same username. 
2. Sensors don't work correctly in this.  Related to DATA_VEHICLE_INSTANCE. When changing this on sensors removes errors from the log but they don't show on the car.  I suspect we need to change this so it is configured to be unique instead of a constant.

### DIFF
--- a/custom_components/kia_uvo/StubTestAPI.py
+++ b/custom_components/kia_uvo/StubTestAPI.py
@@ -16,7 +16,7 @@ from .Token import Token
 
 _LOGGER = logging.getLogger(__name__)
 
-#set class to the one you want to use it for, rename file to that region and make sure it matches.
+# set class to the one you want to use it for, rename file to that region and make sure it matches.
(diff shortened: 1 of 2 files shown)


cdnninja:
@fuatakgun Could you take a look at this?   Currently it works when you add the second car but then fails after you restart the integration.  Reason being is DATA_VEHICLE_INSTANCE is being reset and not stored.   I may need to also rename this variable, treat it more like how you did with PIN.

This however is working for sensors etc on two cars before you restart.   Progress for sure.  To test place the contents of the test file into the hyundai USA API file and add a Hyundai USA car with any username/password combo.

cdnninja:
Closing for now as I am shifting approach.

Issue #196, opened 2021-12-01
GEO Location data is truncated Hyundai USA


rappazzo:
Please check Services, Known Bug / Issues and Troubleshooting over here first: https://github.com/fuatakgun/kia_uvo/blob/master/README.md
**Region and Brand of car**
USA - Hyundai 2020 Ioniq EV

**Describe the bug**
The Location info in the logs seems to have lost its decimal precision
Here is my car's current location ([in the Atlantic Ocean](https://www.latlong.net/c/?lat=40&long=-74)!):
```
         'vehicleLocation': {
            'coord': {
               'alt': 24, 
               'lon': -74, 
               'type': 0, 
               'lat': 40
            }
         }, 
 ```
It looks like this is just truncated (not rounded), as my actual latitude is closer to 41

**To Reproduce**
1. In the lovelace interface, add a map card including the device tracker location of the car
2. Observe the imprecise location.
3. In the logs, observe that the location lat and lon values have no decimal precision

**Expected behavior**
The lat and lon values need a few more decimal places for more accuracy

**Screenshots**
<img width="1012" alt="From Home Assistant" src="https://user-images.githubusercontent.com/525287/144311083-042618b1-aa7f-4698-89c2-f91afb2bbcb3.png">
("mep" is my phone's location)

**Comment by cdnninja** [comment]
Are you able to confirm the Kia app works correctly?

**Comment by rappazzo** [comment]
yeah here is a screen shot (Hyundai app):
![From Hyundai App](https://user-images.githubusercontent.com/525287/144313030-3b503e2e-293d-4ca6-99cb-c8c92afcf8ab.png)
:

**Comment by cdnninja** [comment]
Is the above from the logs or home assistant data? I don't see us doing anything to the location so not sure why this would occur unless they aren't giving us the right amount of accuracy.

**Comment by rappazzo** [comment]
The above data is from the logs (after I turned on debug level logging).  For what it's worth, I used to use Bluelinky with Node Red, and that did have the correct precision in the data.

**Comment by rappazzo** [comment]
Here is the full message from the log: 
```
[...]

PR #201, opened 2021-12-08
fix: Hyundai USA: support a separate location fetch (Issue #196)

rappazzo:
The location information provided in the cached vehicle status does not
include decimal precision.  In order to fix this, implement the
get_location method in the api for Hyundai USA.

Considering that there is a daily API limit, poll the get_location
method based on whether or not the odometer changed.  It will also keep
track of the last time that the location was retrieved form the server,
and only fetch a new location if the last update was more than an hour
ago.


rappazzo:
I am not sure where the commit "fix: black action" came from, but I would rather that be squashed with my original code.

cdnninja:
It’s a bot that corrects style automatically.  

It will squash on merge but you can squash in your branch as well.

rappazzo:
Does this pull request need anything else to be accepted?

PR #205, opened 2021-12-11
fix: Duplicate vehicle.refresh_token() call

cdnninja:
Removed a duplicate vehicle.refresh_token().  This is called in line 195 as refresh_config_entry first refresh's the token.  This is causing duplicate vehicle calls in the USA implementation for Hyundai due to the short token time.


cdnninja:
@rappazzo  I think this may be related to the thread issue you saw.  It is calling refresh token back to back.  I am also seeing some time zone issues that may play a role in it since your refresh is so short.

fuatakgun (approved):

PR #206, opened 2021-12-15
feat: Added support for multiple accounts

cdnninja:
This adds multiple account support.  Only the first car per account though.  This is modeled after dalhb's work.   Config is auto updated from the current build.  I have only tested this once though.  I plan to keep testing but wanted to get a PR in for review.


cdnninja:
This will need squash and merge to ensure it doesn't show so many commits.

custom_components/kia_uvo/__init__.py
```diff
@@ -71,15 +72,30 @@
 
 
 async def async_setup(hass: HomeAssistant, config_entry: ConfigEntry):
-    if DOMAIN not in hass.data:
-        hass.data[DOMAIN] = {}
+    hass.data.setdefault(DOMAIN, {})
+
+    def convert_call_to_vehicle(call) -> Vehicle:
+        vehicle_identifiers = list(hass.data[DOMAIN].keys())
+        if len(vehicle_identifiers) == 1:
```
fuatakgun, line 79: i would get rid of `if` condition here, `else` should work for all conditions.

custom_components/kia_uvo/__init__.py
```diff
@@ -71,15 +72,30 @@
...
+            vehicle_identifier = vehicle_identifiers[0]
+        else:
+            vehicle_identifier = convert_device_id_to_vehicle_identifier(
+                call.data[ATTR_DEVICE_ID]
+            )
+
+        return hass.data[DOMAIN][vehicle_identifier][DATA_VEHICLE_INSTANCE]
+
+    def convert_device_id_to_vehicle_identifier(device_id: str) -> str:
```
fuatakgun, line 88: would rename `get_vehicle_identifier_from_device_id`

custom_components/kia_uvo/__init__.py
```diff
@@ -244,12 +262,29 @@ async def async_unload_entry(hass: HomeAssistant, config_entry: ConfigEntry):
...
 
     return unload_ok
+
+
+async def async_migrate_entry(hass, config_entry: ConfigEntry):
+    if config_entry.version == 1:
+
+        vehicle_id = config_entry.data["stored_credentials"]["vehicle_id"]
+        new_data = config_entry.data.copy()
+        new_data[CONF_VEHICLE_IDENTIFIER] = vehicle_id
```
fuatakgun, line 284: looking here, vehicle_id should be equal to device_id. So, are these equal?
`device_registry.async_get(device_id)` to `vehicle_id`. I personally do not know.

custom_components/kia_uvo/__init__.py
```diff
@@ -244,12 +262,29 @@ async def async_unload_entry(hass: HomeAssistant, config_entry: ConfigEntry):
...
+        ]
         config_update_listener()
 
-        hass.data[DOMAIN] = None
+        hass.data[DOMAIN][vehicle_identifier] = None
 
     return unload_ok
+
+
+async def async_migrate_entry(hass, config_entry: ConfigEntry):
```
fuatakgun, line 279: very nice approach, thanks for saving existing users

custom_components/kia_uvo/config_flow.py
```diff
@@ -52,7 +44,7 @@
 
 
 class KiaUvoOptionFlowHandler(config_entries.OptionsFlow):
-    def __init__(self, config_entry):
+    def __init__(self, config_entry: config_entries.ConfigEntry) -> None:
```
fuatakgun, line 47: 👏

custom_components/kia_uvo/config_flow.py
```diff
@@ -177,3 +181,38 @@ async def async_step_user(self, user_input=None):
         return self.async_show_form(
             step_id="user", data_schema=self.schema, errors=errors
         )
+
+    async def async_step_pick_vehicle(
```
fuatakgun, line 185: are we calling this function?

custom_components/kia_uvo/binary_sensor.py
```diff
@@ -12,14 +12,20 @@
...
+    CONF_VEHICLE_IDENTIFIER,
+)
 
 _LOGGER = logging.getLogger(__name__)
 
 
 async def async_setup_entry(hass, config_entry, async_add_entities):
-    vehicle: Vehicle = hass.data[DOMAIN][DATA_VEHICLE_INSTANCE]
-
+    vehicle: Vehicle = hass.data[DOMAIN][config_entry.data[CONF_VEHICLE_IDENTIFIER]][
```
fuatakgun, line 26: we can add a new field to `KiaUvoEntity.py` class as self.vehicle and all inherited classes will access to `self.vehicle` and we do not need to find correct vehicle under all entities one by one.

custom_components/kia_uvo/device_tracker.py
```diff
@@ -7,13 +7,14 @@
...
 from .KiaUvoEntity import KiaUvoEntity
-from .const import DOMAIN, DATA_VEHICLE_INSTANCE
+from .const import DOMAIN, DATA_VEHICLE_INSTANCE, CONF_VEHICLE_IDENTIFIER
 
 _LOGGER = logging.getLogger(__name__)
 
 
 async def async_setup_entry(hass, config_entry, async_add_entities):
-    vehicle: Vehicle = hass.data[DOMAIN][DATA_VEHICLE_INSTANCE]
+    vehicle_identifier = config_entry.data[CONF_VEHICLE_IDENTIFIER]
```
fuatakgun, line 16: repeat: we can add a new field to `KiaUvoEntity.py` class as self.vehicle and all inherited classes will access to `self.vehicle` and we do not need to find correct vehicle under all entities one by one.

custom_components/kia_uvo/lock.py
```diff
@@ -4,13 +4,19 @@
...
+    VEHICLE_LOCK_ACTION,
+    CONF_VEHICLE_IDENTIFIER,
+)
 
 _LOGGER = logging.getLogger(__name__)
 
 
 async def async_setup_entry(hass, config_entry, async_add_entities):
-    vehicle: Vehicle = hass.data[DOMAIN][DATA_VEHICLE_INSTANCE]
+    vehicle_identifier = config_entry.data[CONF_VEHICLE_IDENTIFIER]
```
fuatakgun, line 18: repeat: we can add a new field to `KiaUvoEntity.py` class as self.vehicle and all inherited classes will access to `self.vehicle` and we do not need to find correct vehicle under all entities one by one.

custom_components/kia_uvo/sensor.py
```diff
@@ -23,13 +23,15 @@
...
     DYNAMIC_DISTANCE_UNIT,
+    CONF_VEHICLE_IDENTIFIER,
 )
 
 _LOGGER = logging.getLogger(__name__)
 
 
 async def async_setup_entry(hass, config_entry, async_add_entities):
-    vehicle: Vehicle = hass.data[DOMAIN][DATA_VEHICLE_INSTANCE]
+    vehicle_identifier = config_entry.data[CONF_VEHICLE_IDENTIFIER]
```
fuatakgun, line 33: repeat: we can add a new field to `KiaUvoEntity.py` class as self.vehicle and all inherited classes will access to `self.vehicle` and we do not need to find correct vehicle under all entities one by one.

custom_components/kia_uvo/services.yaml
```diff
@@ -1,10 +1,36 @@
...
+      name: Device
+      description: Device id of the vehicle.
+      required: false
+      selector:
+        device:
+          integration: kia_uvo
 start_climate:
-  description: TEST - Please use cautiously - start car and climate. Not all items available for all cars.  Refer to Hyundai or Kia app for which are supported for your car. 
+  description: >-
+    TEST - Please use cautiously - start car and climate. Not all items
```
fuatakgun, line 23: let's remove `TEST`, it is well established now, thanks to you

custom_components/kia_uvo/testAPI.py
```diff
@@ -0,0 +1,108 @@
+import logging
```
fuatakgun, line 1: do you want to keep this file here?

fuatakgun (changes requested):
overall looks good, we can squash some more improvements here.

custom_components/kia_uvo/config_flow.py
cdnninja: Not at this point.  That would be used once we have a get_vehicle in all APIs.  Once that is done can select the device and return that in the token.

custom_components/kia_uvo/testAPI.py
cdnninja: Ideally yes.  It is very helpful for testing this new feature so thought I would save the effort from others.

cdnninja:
I will do a round and tweaks.  After running this for a few days I am seeing the occasional error around non unique keys on sensors.   I will have to dig into that. Restarting seems to resolve it.

cdnninja:
For some reason it won't let me reply to the threads anymore.  In response to vehicle id being the same as device id.  Sort of.   Home assistants "device ID" is the car.  Which is our vehicle id.  I think device ID in the token is different and more of a "client id" representing the endpoint from an API standpoint.

PR #211, opened 2021-12-18
fix: Correct 0 value for battery

cdnninja:
@fuatakgun  thoughts on this to ensure the 0 value is not passed?


custom_components/kia_uvo/Vehicle.py
```diff
@@ -73,6 +72,9 @@ async def update(self):
                 _LOGGER.debug(
                     f"zero battery api error, force_update started to correct data"
                 )
+                self.vehicle_data["vehicleStatus"]["evStatus"] = {}
```
fuatakgun, line 75: Not sure how this will impact existing sensors, they will probably generate unavailable, right?

PR #217, opened 2021-12-20
fix(USA Hyundai): Allow weaker ciphers since Hyundai uses these

cdnninja:
Leaving in draft as this only does login right now.

### DIFF
--- a/custom_components/kia_uvo/HyundaiBlueLinkAPIUSA.py
+++ b/custom_components/kia_uvo/HyundaiBlueLinkAPIUSA.py
@@ -25,25 +25,26 @@ from .KiaUvoApiImpl import KiaUvoApiImpl
 from .Token import Token
 
 
-CIPHERS = (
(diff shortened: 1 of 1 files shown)


fuatakgun:
I suggest to do this as a fallback, first try normal more secure way, then on exception, use this less secure way?

cdnninja:
> I suggest to do this as a fallback, first try normal more secure way, then on exception, use this less secure way?

If the server side supports higher I believe it will negotiate up to that.   Other option is I try using a higher cipher default and fall back to lower but that is the same as negotiating the cipher.

cdnninja:
@fuatakgun ready for review as this solves the SSL error on newer version of ubuntu.

cdnninja:
@fuatakgun I reverted the name.  When brackets are used it titles it like this: https://github.com/fuatakgun/kia_uvo/releases/tag/v1.2.4  I do need to watch case in that example.

PR #218, opened 2021-12-21
fix:  Corrected sensor unavailable errors

cdnninja:
Leaving in draft while I watch how my system responds to this over the next few hours.

### DIFF
--- a/custom_components/kia_uvo/__init__.py
+++ b/custom_components/kia_uvo/__init__.py
@@ -237,7 +237,6 @@ async def async_update_options(hass: HomeAssistant, config_entry: ConfigEntry):
 async def refresh_config_entry(hass: HomeAssistant, config_entry: ConfigEntry):
     _LOGGER.debug(f"{DOMAIN} - refresh_config_entry started")
     current_data = config_entry.data.copy()
-    _LOGGER.debug(f"{DOMAIN} - refresh_config_entry current data - {current_data}")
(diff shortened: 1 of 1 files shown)


custom_components/kia_uvo/__init__.py
```diff
@@ -184,12 +184,6 @@ async def async_setup_entry(hass: HomeAssistant, config_entry: ConfigEntry):
 
     async def refresh_config_entry():
```
fuatakgun, line 185: we should not need this function anymore

custom_components/kia_uvo/__init__.py
```diff
@@ -235,6 +229,9 @@ async def async_update_options(hass: HomeAssistant, config_entry: ConfigEntry):
 
 
 async def async_unload_entry(hass: HomeAssistant, config_entry: ConfigEntry):
+    current_data = config_entry.data.copy()
+    current_data[CONF_STORED_CREDENTIALS] = vars(vehicle.token)
+    hass.config_entries.async_update_entry(config_entry, data=current_data)
```
fuatakgun, line 234: do we also need to supply `options` argument too, so they are not lost between saves? not sure what is happening when `options` is missing

fuatakgun (changes requested):
can you please get latest version and give it a try locally?

fuatakgun (approved):

Issue #220, opened 2021-12-24
Set temperature 2°C lower in my car.


zadigre:
I’m in Canada
I have a 2021 Kona Electric

When I start climate, temperature set in the car is 2°C lower.
Temperature shown in HA is same what I’ve set in with start climate. But when I go in my car, it’s 2 lower.

When using the Bluelink app, it’s set properly in my car.

i set it 2 higher in HA and it’s set at my desired temperature.

cdnninja:
I will do another test on my car, assuming it works for me this means the temp range is different on the Hyundai car's vs Kia.  It should be a quick fix.  Thanks for raising it.

dallaby23:
Hi.. first off.. thanks so much for your hard work on this.. I have been using it for awhile now and love the ability to finally start my car.. lol.. I also created a node-red to check for locked then following that command start the vehicle.. works awesome.. 
The reason I jumped in here is I am also in Canada with a Kia Seltos 2021 and when I start_climate with temperature it always sets the temp in the car to 22.5 .. I have tried to start with service set to 24 and 24.5 and 24.0 still always sets it to 22.5 ... Heated steering wheel works fine and front/rear defrost all seem to work fine.. I will try and grab the log and drop it here..

cdnninja:
Hoping to grab time with my wife's car to test this out.  If I can replicate it will be a quick fix. Could I confirm what version you are running?

dallaby23:
1.2.5 .. I just updated before testing this evening..

zadigre:
@cdnninja 
As soon as you release something on the master branch, I’lol test it 😉

cdnninja:
I tested it out.  In my single test using the default values in home assistant my car started but 0.5C off.  I have found a potential bug in the math that impacts 22C only.  Have you tried other temps to see if this impacts multiple temps?  If it is multiple I may need you to capture details from a web browser to see what Kia / Hyundai is passing to start the car.  This could be an electric car issue, I have an ICE.

zadigre:
@cdnninja 
it's happening with multiple temperature for me. 
I usually use 19.5 in Bluelink app. If I try with the same temperature in HA, it will set to 17.5 in the car. I actually do not see it in the car at this temperature (the screen is off), but next time I turn on my car, temperature is at 17.5. 
After finding out about the 2°C difference, I set it to 21.5. and I've got 19.5 in my car.

cdnninja:
[...]

PR #224, opened 2021-12-25
fix(Canada): EV uses different temp range

cdnninja:
This fix only solves start and won't solve get.


cdnninja:
This is for #220  however it sounds like this may be more than just EV.

fuatakgun:
i think this PR is still in draft mode, right?

cdnninja:
Yes need to get year to the api to do this correctly.

Issue #207, opened 2021-12-16
Gaps in data - USA/Hyundai


rappazzo:
**Region and Brand of car**
Hyundai USA

**Describe the bug**
I've been tracking my EV battery in a widget on the UI, and I noticed periods of time in the history where the value is 0 for both short and long periods of time.
![image](https://user-images.githubusercontent.com/525287/146388606-d3f26280-bea8-460a-9b27-755b9ba57209.png)

This seems to come from the server side, so maybe this is a feature request to ignore some responses

**Debug logs if an error occurred**
This is the results from the server with zero'd data for the battery
[zero-batt.txt](https://github.com/fuatakgun/kia_uvo/files/7728055/zero-batt.txt)

**To Reproduce**
1. Track the EV battery status
2. Observe dips and blips in the history where the battery goes from <some value> to 0 and eventually back.  Sometimes it is a  dip, where the zero value lasts a few fetches, and sometimes it is just a blip where it is momentarily zero

**Expected behavior**
At least for an EV, we would *never* expect the battery value to be zero.  Even if zero was a legitimate case, a near-zero low value should suffice.  So, I would like to ignore at least the EV data when the battery value reports 0.  Not sure about other attributes.  It does clearly have values for other attributes (the rough location with truncated decimals as per #196 from the vehicle status response).

fuatakgun:
@dahlb implemented a very clean solution here: https://github.com/fuatakgun/kia_uvo/issues/55
Please double check if problem statement is different.

rappazzo:
I do see:
> zero battery api error, force_update started to correct data
In the logs, but it must still not be triggering in all cases (see the graph above).

cdnninja:
What version number are you running?

rappazzo:
v1.2.3

cdnninja:
Based on the logic I see the potential issue.  It looks like it checks for 0 for current poll and not last poll.   I am wondering if since it is multiple sets of data reporting wrong the system isn't ignoring it. 

Are you seeing that zero battery API message multiple times throughout where it is 0 or just the first poll?

cdnninja:
I think we will need debug log spanning the refreshes to see what is happening here. I am assuming the API just keeps returning a 0 value but would like to confirm.  If so maybe we change the logic to use the last good value and keep trying so other things will refresh. It would show this static and not correct until the API decides to work.

rappazzo:
[...]

PR #233, opened 2022-01-04
fix: use the previous EV state when a blank state is received from the server (#207)

rappazzo:
Maintain the previous vehicle state as a class variable on Vehicle.

### DIFF
--- a/custom_components/kia_uvo/HyundaiBlueLinkAPIUSA.py
+++ b/custom_components/kia_uvo/HyundaiBlueLinkAPIUSA.py
@@ -172,6 +172,14 @@ class HyundaiBlueLinkAPIUSA(KiaUvoApiImpl):
         vehicle_status = {}
         vehicle_status["vehicleStatus"] = response["vehicleStatus"]
 
+        if (
+            vehicle_status["vehicleStatus"]["evStatus"]["batteryStatus"] == 0
+            and HyundaiBlueLinkAPIUSA.old_vehicle_status is not None
+        ):
+            vehicle_status["vehicleStatus"][
+                "evStatus"
+            ] = HyundaiBlueLinkAPIUSA.old_vehicle_status["vehicleStatus"]["evStatus"]
+
         vehicle_status["vehicleStatus"]["dateTime"] = (
             vehicle_status["vehicleStatus"]["dateTime"]
             .replace("-", "")
--- a/custom_components/kia_uvo/Vehicle.py
+++ b/custom_components/kia_uvo/Vehicle.py
@@ -20,6 +20,9 @@ _LOGGER = logging.getLogger(__name__)
 
 
 class Vehicle:
+
+    previous_vehicle_data = None
+
     def __init__(
         self,
         hass: HomeAssistant,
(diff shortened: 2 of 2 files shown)


custom_components/kia_uvo/HyundaiBlueLinkAPIUSA.py
```diff
@@ -172,6 +172,14 @@ def get_cached_vehicle_status(self, token: Token):
         vehicle_status = {}
         vehicle_status["vehicleStatus"] = response["vehicleStatus"]
 
+        if (
+            vehicle_status["vehicleStatus"]["evStatus"]["batteryStatus"] == 0
+            and HyundaiBlueLinkAPIUSA.old_vehicle_status is not None
```
cdnninja, line 177: Should we shift this all to vehicle and remove this from the API?   It looks like the old state is stored in both vehicle and bluelink API.

cdnninja (changes requested):

custom_components/kia_uvo/Vehicle.py
```diff
@@ -51,8 +54,13 @@ def __init__(
 
     async def update(self):
         try:
-            previous_vehicle_status = self.get_child_value("vehicleStatus")
-            previous_vehicle_location = self.get_child_value("vehicleLocation")
+            Vehicle.previous_vehicle_data = self.vehicle_data
```
fuatakgun, line 57: this will create long term issues when moving into `multiple vehicles` as this is a class variable rather than instance one. Can we convert this into instance variable?

custom_components/kia_uvo/Vehicle.py
```diff
@@ -292,8 +300,10 @@ def set_engine_type(self):
                     self.engine_type = VEHICLE_ENGINE_TYPE.IC
         _LOGGER.debug(f"{DOMAIN} - Engine type set {self.engine_type}")
 
-    def get_child_value(self, key):
-        value = self.vehicle_data
+    def get_child_value(self, key, source=None, default=None):
```
fuatakgun, line 303: can we have default valie as `self.vehicle_data`? I do not know if this can be done.

custom_components/kia_uvo/Vehicle.py
```diff
@@ -302,4 +312,6 @@ def get_child_value(self, key):
                     value = value[int(x)]
                 except:
                     value = None
+        if value is None:
```
fuatakgun, line 315: we can write in simpler `value = value or default`

fuatakgun (changes requested):

fuatakgun:
btw, thank you for raising this PR

custom_components/kia_uvo/Vehicle.py
rappazzo: In my experience with using an instance variable with the Hyundai USA implementation I found that there either were multiple instances or that instances were garbage collected and reinitialized.  I mostly inferred this via reading the logs, so I could be wrong, but it seemed like different threads had different instances.

custom_components/kia_uvo/Vehicle.py
rappazzo: not sure what you mean here.

cdnninja:
Closing as this has aged out and V2 this would be a quick fix.   Please raise a PR here if interested: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api

Would think a quick if statement on the value in the API would work in the new model.

Issue #195, opened 2021-12-01
Heated seats constantly report as "On" even when vehicle is switched off


PeteCooney:
**Region and Brand of car**
United Kingdom / Kia (Sorento)

**Describe the bug**
My vehicle has heated front & rear seats. These are reported as 4 sensors (Front Left Seat Heater etc) via the integration, however all 4 constantly report their status as "On" even though they are not. For example; the vehicle is currently parked with the ignition off and they are still showing as "On" in Home Assistant.

**Debug logs if an error occurred**
N/A

**To Reproduce**
1. Enable the integration
2. Sign in
3. Navigate to Configuration >> Devices >> {Vehicle Name}
4. Inspect status of sensor(s)

**Expected behavior**
Sensors should report "Off" when the seat heaters are off and only report "On" when they are switched on.

**Screenshots**
Not really needed, but can provide if required.

**Additional context**
N/A

cdnninja:
Are you able to confirm if this corrects after next scan?

fuatakgun:
Please also share your Data sensor from Developer Tools.

Olivaref:
Hi,
It happens the same to me. 
My car is a Tucson HVE just bought in Spain. The Bluelink app reports all the seats heaters are off as these actually are, but HA reports all on (Encendido).
Here is a screenshot.
![Captura de pantalla 2021-12-01 182417](https://user-images.githubusercontent.com/31900913/144283031-14363cc9-8854-47ab-b45b-bb09002e312f.png)

cdnninja:
Are you able to turn them on, force a update and post your data again? The data shows a "2" maybe that means off and a different number means on?  Please also try turning the car off while heated seats are on and see if the value changes from whatever "on" is.

cdnninja:
As well what does the Kia app report?

Olivaref:
I've done a quick test. The data I sent previously was with the car's ignition off.
Now, I started the car, and switched on the driver's seat heater to max power (Three orange leds lighting), and switched on the copilot seat air conditioner/fan to min power (One blue led lighting). 
In this Tucson, seats can be heated or cooled with three power levels. The first picture shows the switches for both front seats for heating and cooling with the orange and blue leds for the test condition. 
I saw the status changed in the Bluelink app (Bluelink picture), but HA reminded unchanged (value 2 for those sensors, HA picture). I made sure HA was updated.   
[...]

PR #245, opened 2022-01-12
feat: Added Seat Status

cdnninja:
Updated to reflect correct states in sensors.  Assuming all regions are the same as we have had two reports from different regions that align.


cdnninja:
I have not tested this yet.   Would resolve #195.

cdnninja:
I am thinking we merge as is and see what feedback we get.  Thoughts?

fuatakgun (approved):

PR #246, opened 2022-01-13
fix: force update scan when start < finish

rsegers:
This PR fixes the issue when `no_force_scan_hour_start` is smaller than `no_force_scan_hour_finish`.
For example, configuring no force update between 01h00 and 05h00 would result in the forced update _only_ triggering between those hours, in stead of outside the range. 

The `if`-statement now checks whether `no_force_scan_hour_start <= no_force_scan_hour_finish` and applies a slightly different logic.


cdnninja (approved):

cdnninja:
Thanks for the contribution!  Greatly appreciated.

PR #251, opened 2022-01-15
feat: Added Seat Status

cdnninja:
Keeping in draft while we address issues.


fuatakgun (commented):
~~I don't understand what is difference now, is this working locally?~~

fuatakgun:
> Keeping in draft while we address issues. 

My bad, this is draft

PR #260, opened 2022-01-21
fix: Added HACS minimum HA Version

cdnninja:
Sets the minimum home assistant version required for this to run as we have seen two issues reported due to this.


fuatakgun:
That is something cool. Just one clarification;
- HA: home assistant
- HACS: home assistant custom store

cdnninja:
Ahh yes. The pr title is wrong.

cdnninja:
Yes it was an edit to the hacs file to evidence a min ha version.

PR #263, opened 2022-01-23
feat: Location Entity

cdnninja:
Still requires a bit of API work for this value to exist in the car.    Will post for review purposes. Won't merge until I have the API working on this as well.

### DIFF
--- a/custom_components/kia_uvo/device_tracker.py
+++ b/custom_components/kia_uvo/device_tracker.py
@@ -27,13 +27,13 @@ async def async_setup_entry(
     entities = []
     for vehicle_id in coordinator.vehicle_manager.vehicles.keys():
         vehicle: Vehicle = coordinator.vehicle_manager.vehicles[vehicle_id]
-        entities.append(HyundaiKiaConnectLocationTracker(coordinator, vehicle))
+        entities.append(HyundaiKiaConnectTracker(coordinator, vehicle))
 
     async_add_entities(entities)
     return True
 
 
-class HyundaiKiaConnectLocationTracker(TrackerEntity, HyundaiKiaConnectEntity):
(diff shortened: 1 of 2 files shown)


cdnninja:
Should icon, and source type move to attributes since they don't change?

custom_components/kia_uvo/device_tracker.py
```diff
@@ -0,0 +1,58 @@
...
+    @property
+    def latitude(self):
+        return self.vehicle.location_latitude
+
+    @property
+    def longitude(self):
+        return self.vehicle.location_longitude
+
+    @property
+    def source_type(self):
```
fuatakgun, line 57: Nit: let's use attr field

custom_components/kia_uvo/device_tracker.py
```diff
@@ -0,0 +1,58 @@
...
+    entities = []
+    for vehicle_id in coordinator.vehicle_manager.vehicles.keys():
+        vehicle: Vehicle = coordinator.vehicle_manager.vehicles[vehicle_id]
+        entities.append(HyundaiKiaConnectLocationTracker(coordinator, vehicle))
+
+    async_add_entities(entities)
+    return True
+
+
+class HyundaiKiaConnectLocationTracker(TrackerEntity, HyundaiKiaConnectEntity):
```
fuatakgun, line 36: nit: HyundaiKiaConnectTracker or HyundaiKiaConnectDeviceTracker

fuatakgun (changes requested):
Perfect

custom_components/kia_uvo/device_tracker.py
cdnninja: https://developers.home-assistant.io/docs/core/entity/device-tracker/ states the following so I don't think I can. 

`CAUTION
TrackerEntity does not support attribute shorthand for property implementation`

PR #270, opened 2022-01-28
fix: last_updated unit of measurement

rsegers:
Super minor fix for the unit_of_measurement for last_update. Should not be the string `"None"` but simply state `None` (just like geocodedLocation)

It fixes my InfluxDB measurements 👍


cdnninja:
Thank you for the contribution!  Great catch.

Issue #275, opened 2022-02-04
Force update not working.


Growiel:
Please check Services, Known Bug / Issues and Troubleshooting over here first: https://github.com/fuatakgun/kia_uvo/blob/master/README.md
**Region and Brand of car**

Ioniqu Electric v2, EU

**Describe the bug**

`force_update` doesn't work anymore.

**Debug logs if an error occurred**

```
kia_uvo - Exception in update : string indices must be integers - traceback: Traceback (most recent call last): File "/config/custom_components/kia_uvo/Vehicle.py", line 56, in update self.vehicle_data = await self.hass.async_add_executor_job( File "/usr/local/lib/python3.9/concurrent/futures/thread.py", line 52, in run result = self.fn(*self.args, **self.kwargs) File "/config/custom_components/kia_uvo/KiaUvoApiEU.py", line 427, in get_cached_vehicle_status return response["resMsg"]["vehicleStatusInfo"] TypeError: string indices must be integers
```

I don't really have more info to share, could be due to an outage with the API, my Bluelink app is also not working, but if that's the case, I guess we need a better way to handle it (show an error message ?)

EDIT: Both the Bluelink app and the integration are back.

cdnninja:
Could you test my code form 276 and let me know how it handles the next failure?  It may just shift the error further up.

Growiel:
Here are the results, it ws not working again this morning.

`Logger: homeassistant.components.websocket_api.http.connection
Source: custom_components/kia_uvo/__init__.py:79
Integration: Home Assistant WebSocket API (documentation, issues)
First occurred: 7:19:52 AM (1 occurrences)
Last logged: 7:19:52 AM

[140577562747232] 'vehicle'
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 190, in handle_call_service
    await hass.services.async_call(
  File "/usr/src/homeassistant/homeassistant/core.py", line 1630, in async_call
    task.result()
  File "/usr/src/homeassistant/homeassistant/core.py", line 1667, in _execute_service
    await cast(Callable[[ServiceCall], Awaitable[None]], handler.job.target)(
  File "/config/custom_components/kia_uvo/__init__.py", line 79, in async_handle_force_update
    vehicle: Vehicle = hass.data[DOMAIN][DATA_VEHICLE_INSTANCE]
KeyError: 'vehicle'
`

fuatakgun:
[...]

PR #276, opened 2022-02-04
fix(EU): Only return data when it exists

cdnninja:
This is related to #275.  Not sure how the parent calling method will handle this though.


custom_components/kia_uvo/KiaUvoApiEU.py
```diff
@@ -424,7 +424,13 @@ def get_cached_vehicle_status(self, token: Token):
...
         response = response.json()
         _LOGGER.debug(f"{DOMAIN} - get_cached_vehicle_status response {response}")
-        return response["resMsg"]["vehicleStatusInfo"]
+        if response["resCode"] == 0000:
+            return response["resMsg"]["vehicleStatusInfo"]
+        else:
+            _LOGGER.debug(
+                f"{DOMAIN} - get_cached_vehicle_status API failed to respond with data"
+            )
+            return None
```
fuatakgun, line 433: returning None would generate similar error in upstream functions, will it? I might be missing something.

PR #277, opened 2022-02-05
fix(EU): Call location details

cdnninja:


custom_components/kia_uvo/KiaUvoApiEU.py
```diff
@@ -443,6 +444,31 @@ def get_geocoded_location(self, lat, lon):
         response = response.json()
         return response
 
+    def get_location(self, token: Token):
```
fuatakgun, line 447: cached vehicle status is just getting data from cloud without forcing the car to refresh its data. `get_location` will require car to refresh itself, we should follow a similar principle you have delivered in CA, if odometer has changed (or sth else) and vehicle data time is out sync location data time, we should trigger this. Sharing my example data below;

```
vehicle_data:
  vehicleLocation:
    coord:
      lat: xxx
      lon: yyy
      alt: 4
      type: 0
    head: 253
    time: '20220206132045'
  vehicleStatus:
    time: '20220206132047'
  odometer:
    value: 20388.5
    unit: 1
vehicle_name: NIRO
friendly_name: NIRO Data

```

Vehicle location time is 2 seconds earlier than vehicle data time, so this should be fine, to some extent. But sometimes, for some awkward reasons, they get out of sync and we should call vehicle/location. I would just call this function if we are out of sync by x timedelta and call another get cached vehicle status.

cdnninja: Perfect.  This is exactly the type of data we needed to solve this.    Could I also get the details from logs on what is contained in the get location response?  Does it have the same coordinates or just a success so we have to grab it from cached vehicle status?   If it contains the location we can use its values instead.

fuatakgun: output of `location` api

```
{
  'retCode': 'S',
  'resCode': '0000',
  'resMsg': {
    'gpsDetail': {
      'coord': {
        'lat': 40.123456,
        'lon': 1.23456,
        'alt': 4,
        'type': 0
      },
      'head': 254,
      'speed': {
        'value': 0,
        'unit': 0
      },
      'accuracy': {
        'hdop': 10,
        'pdop': 10
      },
      'time': '20220209132104'
    },
    'drvDistance': {
      'rangeByFuel': {
        'gasModeRange': {
          'value': 390,
          'unit': 1
        },
        'evModeRange': {
          'value': 58,
          'unit': 1
        },
        'totalAvailableRange': {
          'value': 448,
          'unit': 1
        }
      },
      'type': 2
    }
  },
  'msgId': '381d8f6f-c0b8-45ea-9f36-6530298eb5a9'
}
```

cdnninja: I have placed this in the force call for now.   To do the canadian odometer fix I have to start storing vehicle state within the call. I didn't like this for canada but this is solved in the new API version so doing a quick fix for now.

cdnninja: @fuatakgun  Could you test this for a bit on your instance?

custom_components/kia_uvo/KiaUvoApiEU.py
```diff
@@ -443,6 +443,27 @@ def get_geocoded_location(self, lat, lon):
         response = response.json()
         return response
 
+    def update_location(self, token: Token):
+        url = self.SPA_API_URL + "vehicles/" + token.vehicle_id + "/location"
+        headers = {
+            "Authorization": token.access_token,
+            "Stamp": token.stamp,
```
Lerbaek, line 450:
```suggestion
            "Stamp": self.get_stamp(),
            "ccsp-service-id": self.CCSP_SERVICE_ID,
            "ccsp-application-id": self.APP_ID,
```

Lerbaek (changes requested):
The suggested changes have been tested working on my system with v1.5.0

custom_components/kia_uvo/KiaUvoApiEU.py
Lerbaek: I'm not certain that all these headers and the change are needed, but this aligns the request with the one from `update_vehicle_status` below.

cdnninja: I don't see get_stamp() as a function?

Lerbaek: `get_stamp()` was introduced in 6fec63beadcdb07bebea6351773f1ea3c5b7e29d on Feb 8, so you'll need to bring this branch up to date for my suggestion to be applicable.

cdnninja: Good catch. Thank you!

Lerbaek:
Just did a file comparison, and the file is now identical to the one with which I've tested, as mentioned in in #47. ✅

Lerbaek (approved):
Just did a file comparison, and the file is now identical to the one with which I've tested, as mentioned in in https://github.com/fuatakgun/kia_uvo/issues/47.

Lerbaek:
So how are we doing on this, and can I or anyone else do anything to help get this merged? Do you need more test pilots?

cdnninja:
I will merge.  Hoping to hear if this causes issue or solves the issue.

PR #286, opened 2022-02-07
fix(EU): Corrected token

cdnninja:
@fuatakgun  take a look to see if this makes sense.  Looks like some are reporting it fails after a bit which would make sense on force update.


fuatakgun:
access token and refresh token are different, this would break EU use cases.

PR #292, opened 2022-02-09
feat(US Kia): Added miles until next service information

nprez83:
Update API request to download maintenance information and added it to vehicle_data to be returned as a sensor entity (sensor is already defined in sensor.py, variable just needed to be defined within KiaUvoAPIUSA file).

### DIFF
--- a/custom_components/kia_uvo/KiaUvoAPIUSA.py
+++ b/custom_components/kia_uvo/KiaUvoAPIUSA.py
@@ -289,6 +289,7 @@ class KiaUvoAPIUSA(KiaUvoApiImpl):
                     ]["vehicle"]["mileage"]
                 ),
                 "unit": 3,
+            },
             "nextService": {
                 "value": float(
                     response_body["payload"]["vehicleInfoList"][0]["vehicleConfig"][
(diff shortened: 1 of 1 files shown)


cdnninja:
Awesome! Thank you for the contribution!   With this change do you see nextservice sensor?

nprez83:
> Awesome! Thank you for the contribution! With this change do you see nextservice sensor in testing?

Indeed. With those changes, the nextservice entity sensor is loaded automatically, because it was already defined in the sensors.py file, so all we needed to do was pull the info in the API request, and then return it. Now there are 24 entities linked to the device, instead of 23 previously.

cdnninja:
Awesome.  I think in CA we have a last service sensor too that you could map.  In the meantime I will merge.  Thanks again! Releases occur wednesday and weekends so until that time it will need to be updated to "latest" to have this.

nprez83:
Awesome, thanks. I just looked at the log, and it looks like all the maintenance info I'm getting is the schedule and miles till next service (see below). I think @dahlb is using the odometer reading, appending it to the maintenance array, and indexing over that to figure out the last and next service, which is pretty clever. TBH, I'm satisfied just having the miles till next service, which can be used to trigger alerts. Thanks again!

`"maintenance":{"nextServiceMile":3112.5996,"maintenanceSchedule":[7500,15000,22500,30000,37500,45000,52500,60000,67500,75000,82500,90000,97500,105000,112500]}`

PR #293, opened 2022-02-09
fix(US): Correct temperature settings to match the kia UVO app and the vehicle

nprez83:
In the US Kia Telluride (and I assume other US Kia vehicles) as well as the US Kia UVO app, you are only allowed to set the climate temperature to LOW -> 62-82 <- HIGH. I made changes to the services.yaml file to account for this: 1) Changed the default temperature from 21 to 72, 2) The min to 60, 3) And the max to 85. I also added logic in the vehicle.py file to translate anything < 62 to 'LOW', and anything > 82 to 'HIGH'.

### DIFF
--- a/custom_components/kia_uvo/Vehicle.py
+++ b/custom_components/kia_uvo/Vehicle.py
@@ -181,7 +181,11 @@ class Vehicle:
 
     async def start_climate(self, set_temp, duration, defrost, climate, heating):
         if set_temp is None:
-            set_temp = 21
+            set_temp = 72
+        elif set_temp < 62: 
+            set_temp = 'LOW'
+        elif set_temp > 82:
+            set_temp = 'HIGH'
         if duration is None:
             duration = 5
         if defrost is None:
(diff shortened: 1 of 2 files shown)


cdnninja:
This will break celsius users which is the majority of the users of the integration.   Services.yaml can't have much logic so thinking that needs to be left pretty open but the default call that you are editing could check the unit or region potentially.

nprez83:
I see, totally get it. How about this solution? I added just a small note to the services.yaml file letting users know that if they have a US Kia, anything <62 will translate to "LOW" and anything >82 will translate to "HIGH", as it is in the vehicle itself and the UVO app. No other changes to the services.yaml file. I then added an if statement to the Vehicle.py file to test for the users region, and if it's US to apply the logic, otherwise things will be as they were previously. 

I probably should have mentioned before by the way, but the reason I'm trying to change this is because in the US Kia integration, if you try to set the climate to a number lower than 62 or higher than 85, it causes an error that leads to the engine and the rear defroster coming on, but the AC does not, which is why adding the logic check to the Vehicle.py file is actually a needed fix to prevent this error.

cdnninja:
Looks good.  Items like this will also be easier in the 2.0 version in the works.   I tagged you in the last backport.

PR #297, opened 2022-02-11
feat: Remote Start

cdnninja:
Draft for feedback while the logic is completed.


cdnninja:
Graphics of what it looks like: 
image
image
image

The commands are complete to support this in the PR just yet but I will keep working on that while reviewed.

custom_components/kia_uvo/services.py
```diff
@@ -49,6 +50,7 @@ async def async_handle_update(call):
 
     async def async_handle_start_climate(call):
         coordinator = _get_coordinator_from_device(hass, call)
+        climate_request_options = ClimateRequestOptions(set_temp = call.data["temperature"], duration = call.data["duration"], climate = call.data["climate"], heating = call.data["heating"])
```
fuatakgun, line 53: if we have any intention to default any of these inputs, we can do here as call.data.get("variable","default")

fuatakgun (approved):
Looks good to me

PR #305, opened 2022-02-18
feat: batteryPreconditioning as a sensor

henrikbjorn:


cdnninja:
Thanks for the contribution!

PR #306, opened 2022-02-24
feat: add climate control as climate platform

jwefers:
Adding climate control to the HYU/KIA integration.

This is a total Draft. My Ioniq Electric is only arriving in a few weeks and i never touched python for anything bigger than a 10-liner.

However, i would like my car to be fully controllable so i want to devote time into the HA integration. Since you guys already started the process, i would like to append to your work :)

Please let me know your thoughts and code criticisms since you also build the actual HYU/KIA Connect python lib.

**As discussed in https://github.com/fuatakgun/core/pull/7, i understand basing this on top of `generic_thermostat` might make more sense. However before i can approach that:**

- i need a car, hopefully 1-2 weeks :laughing: 
- generic_termostat wants the toggles for heater/ac be separate switches. It also treats heater/ac as ONE toggle based on ac mode. Once i have a test bench, i can see if that mapping can be done sensibly.

### DIFF
--- a/custom_components/kia_uvo/__init__.py
+++ b/custom_components/kia_uvo/__init__.py
@@ -22,6 +22,7 @@ PLATFORMS: list[str] = [
     Platform.SENSOR,
     Platform.DEVICE_TRACKER,
     Platform.LOCK,
+    Platform.CLIMATE,
 ]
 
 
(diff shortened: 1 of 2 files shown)


jwefers:
@cdnninja @fuatakgun  This code now actually works - with a european Ioniq Electric. Would you test with yours cars as well? There is still a lot of rough edges though, where i have to learn the HA concepts and improve.

PR #310, opened 2022-03-20
feat(EU): set charging limits as number_input

jwefers:
NOTE: this depends on
- https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/53
- https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/54

After the lib provides the features to get/set EV charging limits, this provides them as number integration for EU:
image

### DIFF
--- a/custom_components/kia_uvo/number.py
+++ b/custom_components/kia_uvo/number.py
@@ -1,0 +1,84 @@
+"""Device Tracker for Hyundai / Kia Connect integration."""
+from __future__ import annotations
+
+import logging
+from typing import Final
+
+from hyundai_kia_connect_api import Vehicle, VehicleManager
+
+from homeassistant.components.number import NumberEntity, NumberEntityDescription
+from homeassistant.config_entries import ConfigEntry
+from homeassistant.const import ATTR_TEMPERATURE
+from homeassistant.core import HomeAssistant
+from homeassistant.helpers.entity_platform import AddEntitiesCallback
+
+from .const import DOMAIN
+from .coordinator import HyundaiKiaConnectDataUpdateCoordinator
+from .entity import HyundaiKiaConnectEntity
+
+_LOGGER = logging.getLogger(__name__)
+
+CHARGING_LIMIT_DESCRIPTIONS: Final[tuple[NumberEntityDescription, ...]] = (
+    NumberEntityDescription(
+        key="_ac_charging_limit",
+        name="AC Charging Limit",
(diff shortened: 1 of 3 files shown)


custom_components/kia_uvo/number.py
```diff
@@ -0,0 +1,119 @@
...
+        """Set new charging limit."""
+        # force refresh of state so that we can get the value for the other charging limit
+        # since we have to set both limits as compound API call.
+        current_limits = await self.hass.async_add_executor_job(
+            self._vehicle_manager.api.get_charge_limits,
+            self._vehicle_manager.token,
+            self.vehicle,
+        )
+
+        # don't do anything for null change
```
jwefers, line 93: here is a bug - when setting the limits on a different channel (e.g. bluelink app) and then pulling the slider in HA to the same target percentage, it stops here and the UI will stay at the old displayValue - until the configured refresh cycle finally hits. will fix soon

custom_components/kia_uvo/number.py
```diff
@@ -0,0 +1,119 @@
...
+        if (
+            self.entity_description.key == AC_CHARGING_LIMIT_KEY
+            and current_limits.ac == int(value)
+        ):
+            return
```
jwefers, line 98: here is a bug - when setting the limits on a different channel (e.g. bluelink app) and then pulling the slider in HA to the same target percentage, it stops here and the UI will stay at the old displayValue - until the configured refresh cycle finally hits. will fix soon

Growiel: @jwefers were you ever able to get to fixing this ? I tried it and was able to set the charge limits and couldn't really recreate the issue. Thanks.

cdnninja:
I am going to merge this however it is to the version 2.0 branch that needs testers and work to get to feature parity.  Low risk merge in my view.

custom_components/kia_uvo/number.py
jwefers: No, i did not find the time since March to work on this :D the bug is obviously still present - eventually i need to rethink myself into it. I think what needs to be done here is to consider the internally cached state with what the API returns AND the desired number.

rindlerblabla:
Am interested in this feature, but it's not there on 2.0? Where did it go, hehe?

cdnninja:
It was merged. Must be a bug on filtering data as I did have to tweak it.   Could you open a new issue for this?

PR #313, opened 2022-04-12
feat(EU): add average daily electric consumption sensor

tobiasbayer:
I only implemented this for EU as I do not know if the other regions support the same type of drive history.


tobiasbayer:
This use case can be handled with https://github.com/fuatakgun/kia_uvo/pull/308 and utility meters with daily cycles on the odometer and the monthly total usage.

PR #315, opened 2022-04-12
fix: do not translate title

tobiasbayer:
triggers hassfest validation


cdnninja:
Thank you for the contributions!

PR #347, opened 2022-08-15
fix(USA Hyundai): doorLockStatus now uses doorLock

GD-L:
References the correct vehicle attribute with boolean response.
Changed the response reference from `doorLockStatus` to `doorLock`.


cdnninja:
Thank you for the contribution! I assume this has tested successfully so I will merge.

Issue #370, opened 2022-10-06
Unchecked charge limit sets limit to 90%


richy58729:
**Region and Brand of car**
Region is EU, brand is Kia, model is e-Niro MY22

**Describe the bug**
When using the set_charge_limits service (BTW, thanks for this service, opens up new possibilities!), an unchecked limit (doesn't matter if it's DC or AC) sets the limit in the UI to 90%, but after calling the service, the unchecked charge limit is also set to 90% in reality. If you uncheck both limits (which is of course nonsense if you don't check anything, but I did it just for the sake of testing), both limits will be set to 90% in reality (checked in the Kia Connect app BTW, not in the car itself).

**To Reproduce**
1. Go to 'Developer Tools &rarr; Services'
1. Look up the service 'kia_uvo: set_charge_limits'
1. Make sure DC or AC is unchecked (or both are)
1. Click on 'Call Service'
1. Look at 'Set charging limits' in the Kia Connect app and the unchecked limit is set to 90%, or use the 'kia_uvo: update' service and see the targetSOClevel for plugType 0 or 1 (or both) is set to 90.

**Expected behavior**
An unchecked limit should not change the limit in reality.

cdnninja:
I don't have an EV so makes it hard to confirm.  Could you check if the mobile app lets you set a single one of these?  If it shows both regardless does it just default the value to what it currently is set to?

richy58729:
As soon as I tap _Set charging limits_ in the Kia Connect app,
![Screenshot_2022-10-07-06-44-41-14_8c1be1e5aae6d26dc95284741c49c73b](https://user-images.githubusercontent.com/1909971/194773453-186cbc61-aaf3-44e9-870d-0e5e4c6e3ccd.jpg)

the app queries the current charging limits. and then I see the current limits, which I can change and tap _Send to car_.
![Screenshot_2022-10-07-06-44-58-03_8c1be1e5aae6d26dc95284741c49c73b](https://user-images.githubusercontent.com/1909971/194773469-82462418-cb2f-4f0d-8429-759bf2812f0c.jpg)

So no, you can't set a single limit, but both. And yes, it defaults to the current charging limits.

cdnninja:
In the home assistant gui we can't set dynamic defaults.   So how I am thinking we handle this is when the box is unchecked query for the current value and pass that.   If both are unchecked just cancel the request.   

Thoughts?

richy58729:
# UI
[...]

PR #377, opened 2022-10-15
fix: Charge limit logic change

cdnninja:
The start of logic to not use defaults to set a charge limit.  Goal will be to get current limits rather than set defaults.


cdnninja:
Resolves #370

cdnninja:
I have not tested this in any way as I don't have an EV.

PR #406, opened 2022-11-03
fix: Don't use device_class battery for 12V battery (master)

### LINKED ISSUE #240: Use "EV Battery" for device instead of "Car Battery" for EVs
I think the title says it all.

When using the integration with an EV such as IONIQ 5, the device shows battery as 88% which is the 12v battery and not the EV battery, which is way more interesting for EV cars.

hmmbob:
SSIA.
As per discussion with @cdnninja 

Fixes #240


cdnninja:
Thank you for this!

Issue #411, opened 2022-11-05
Charging limit not working anymore


rindlerblabla:
EU EV6

**Describe the bug**
Since 2.0 it is not possible to set charging limit anymore.

> 
Logger: homeassistant.helpers.script.websocket_api_script
Source: custom_components/kia_uvo/coordinator.py:137
Integration: Kia Uvo / Hyundai Bluelink (documentation, issues)
First occurred: 15:02:17 (2 occurrences)
Last logged: 15:02:24

websocket_api script: Error executing script. Unexpected error for call_service at pos 1: VehicleManager.set_charge_limits() takes 3 positional arguments but 4 were given
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 451, in _async_step
    await getattr(self, handler)()
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 684, in _async_call_service_step
    await service_task
  File "/usr/src/homeassistant/homeassistant/core.py", line 1738, in async_call
    task.result()
  File "/usr/src/homeassistant/homeassistant/core.py", line 1775, in _execute_service
    await cast(Callable[[ServiceCall], Awaitable[None]], handler.job.target)(
  File "/config/custom_components/kia_uvo/services.py", line 98, in async_handle_set_charge_limit
    await coordinator.set_charge_limits(
  File "/config/custom_components/kia_uvo/coordinator.py", line 137, in set_charge_limits
    await self.hass.async_add_executor_job(
  File "/usr/local/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
TypeError: VehicleManager.set_charge_limits() takes 3 positional arguments but 4 were given

cdnninja:
Could you retry on the latest?

rindlerblabla:
Unfortunately it is still not working. Causing error:

Logger: homeassistant.helpers.script.websocket_api_script
Source: custom_components/kia_uvo/services.py:100
Integration: Kia Uvo / Hyundai Bluelink (documentation, issues)
First occurred: 07:51:52 (2 occurrences)
Last logged: 07:53:17

websocket_api script: Error executing script. Unexpected error for call_service at pos 1: HyundaiKiaConnectDataUpdateCoordinator.set_charge_limits() missing 1 required positional argument: 'dc_limit'
Traceback (most recent call last):
[...]

PR #412, opened 2022-11-05
fix:  Set Charge Limits

cdnninja:
Corrected ev charge limits service.


cdnninja:
This will resolve #411 However I think the underlying library needs the evlimits datatype moved before we test this out.

PR #421, opened 2022-11-07
fix: Number alignment to coordinator model

cdnninja:
This is untested as I don't have an EV.


jwefers:
breaks setting the charge limits - the code somehow never returns from async_force_update_all. I will investigate a bit

jwefers:
i see many many errors while debugging through it unfortunately. Most notably that you renamed the references from `_vehicle` to `vehicle`, but did not rename the class variable - leading to undeclared variable errors. Then i also found in another PR you changed the  flow in how force refresh was implemented and it also does not match to what the API returns.

I am gonna try to get a compound PR done, but the errors are in both the library and here :(

cdnninja:
Thank you! Appreciated.  I should have held off merging the initial addition instead of trying to align to coordinator model after the fact.  

This is a beta release so feel free to drop smaller PR if it makes life easier.

PR #444, opened 2022-11-17
fix: Charging Number Entity

cdnninja:


cdnninja:
@jwefers  Did a little bit of tinkering.  Interested to see thoughts and errors.  Feel free to drop a separate PR if you have something in the works.  Was reading your feedback on the other thread and thought I would move it here. 

This is trying to match the model in https://github.com/fuatakgun/kia_uvo/blob/v2.0/custom_components/kia_uvo/sensor.py
and the commands used in https://github.com/fuatakgun/kia_uvo/blob/v2.0/custom_components/kia_uvo/lock.py

This will allow other items to move into this class in the future as well, if it makes sense.  

Currently I think it won't show up because of the way evlimits is setup.  This protects it from erroring on ICE vehicles.

zadigre:
Just to let you know... I confirm nothing (more) shows up in the entities list after the upgrade to 2.0.37

cdnninja:
Yes, still needs some data changes.  I have a pr on the api side I'm working from that fixes the data issues.

Issue #440, opened 2022-11-15
Last Action Tracking


cdnninja:
Implement last action tracking in 2.0.

torbjornaxelsson:
Hey guys, I'm working on a solution for action tracking, see pull request linked above. Thankful for your thoughts! I've tested it in my own Home Assistant setup and it seems to work.

PR #511, opened 2022-12-17
feat: Check for updates WIP

torbjornaxelsson:
Proposal for tracking API status updates, solution for issue #440. Also see pull request in the hyundai_kia_connect_api repo with required changes to the API. Work in progress, thankful for any input. The code works for me (CA/Hyundai)

**Changes in kia_uvo**
* all vehicle commands call async_wait_for_action_completed after the action has beeen sent (rather than async_request_refresh which just receives stale data)
* async_wait_for_action_completed calls vm.check_action_status every X seconds a maximum of Y times
* when vm.check_action_status responds True, async_refresh() is called to refresh values in HA

**Changes in hyundai_kia_connect_api**
* CA.check_last_action_status updates vehicle properties if action completed successfully
* Token gets parameter last_action_pin_token required for CA.check_last_action_status
* EU gets a mock check_last_action_status for compatibility

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -179,65 +179,61 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
 
     async def async_lock_vehicle(self, vehicle_id: str):
         await self.async_check_and_refresh_token()
-        action_id = await self.hass.async_add_executor_job(
-            self.vehicle_manager.lock, vehicle_id
-        )
-        await self.async_wait_for_action_completed(vehicle_id, action_id)
+        await self.hass.async_add_executor_job(self.vehicle_manager.lock, vehicle_id)
+        await self.async_wait_for_action_completed(vehicle_id)
 
     async def async_unlock_vehicle(self, vehicle_id: str):
         await self.async_check_and_refresh_token()
-        action_id = await self.hass.async_add_executor_job(
-            self.vehicle_manager.unlock, vehicle_id
-        )
-        await self.async_wait_for_action_completed(vehicle_id, action_id)
+        await self.hass.async_add_executor_job(self.vehicle_manager.unlock, vehicle_id)
+        await self.async_wait_for_action_completed(vehicle_id)
 
     async def async_open_charge_port(self, vehicle_id: str):
         await self.async_check_and_refresh_token()
-        action_id = await self.hass.async_add_executor_job(
+        await self.hass.async_add_executor_job(
             self.vehicle_manager.open_charge_port, vehicle_id
         )
-        await self.async_wait_for_action_completed(vehicle_id, action_id)
+        await self.async_wait_for_action_completed(vehicle_id)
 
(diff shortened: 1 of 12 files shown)


custom_components/kia_uvo/coordinator.py
```diff
@@ -159,63 +160,83 @@ async def async_check_and_refresh_token(self):
             self.vehicle_manager.check_and_refresh_token
         )
 
+    async def async_wait_for_action_completed(
+        self, vehicle_id: int, action_id: str, interval: int = 5, max_count: int = 6
+    ):
+        for count in range(1, max_count + 1):
+            _LOGGER.debug(f"Last action check: waiting {interval} seconds")
+            await asyncio.sleep(interval)
```
cdnninja, line 168: I don't have time to check right now but I believe home assistant discourages this as it ties up the event loop.  They have a specific command that you use to have it schedule something X time in the future instead.

custom_components/kia_uvo/coordinator.py
```diff
@@ -159,63 +160,83 @@ async def async_check_and_refresh_token(self):
...
+            if is_completed:
+                await self.async_refresh()
+                break
+        return is_completed
+
     async def async_lock_vehicle(self, vehicle_id: str):
         await self.async_check_and_refresh_token()
-        await self.hass.async_add_executor_job(self.vehicle_manager.lock, vehicle_id)
-        await self.async_request_refresh()
+        action_id = await self.hass.async_add_executor_job(
```
cdnninja, line 181: With action ID stored in the coordinator this won't work for multiple cars as it will be overwritten.  So maybe we store the action ID within vehicle in the API and purely call check last action without the ID?

cdnninja (changes requested):

cdnninja:
Thank you for the work on this! Greatly appreciate it!

custom_components/kia_uvo/coordinator.py
torbjornaxelsson: Good catch! I realized we may have the same issue with last_pauth so now storing them both per vehicle

custom_components/kia_uvo/coordinator.py
torbjornaxelsson: Thank you, will see if I can figure out the correct way to do it

torbjornaxelsson: Okay, so I gave this some thought. I wonder if the polling logic belongs in the API rather than the coordinator? Given that we support checking status for multiple vehicles simultaneously, and polling logic may be API-dependent.

Not sure this is the best way to go about it, but we could add a callback to vehicle_manager.check_action_status that gets executed when the action is completed. In the coordinator we could then have this callback function refresh home assistant data, handle errors etc.

On the API side we could poll using asyncio.call_later for or something similar to avoid locking the event loop.

Would appreciate your thoughts on this!

kalaws:
Is this PR about to be implemented?

cdnninja:
No.  The parent pr had feedback that hasn't been addressed.

PR #559, opened 2023-01-07
feat: set SOC levels as number instead of slider

bouwew:
This update will improve the usability of the numbers:
image

### DIFF
--- a/custom_components/kia_uvo/number.py
+++ b/custom_components/kia_uvo/number.py
@@ -6,7 +6,11 @@ from typing import Final
 
 from hyundai_kia_connect_api import Vehicle, VehicleManager
 
-from homeassistant.components.number import NumberEntity, NumberEntityDescription
+from homeassistant.components.number import (
+    NumberEntity,
+    NumberEntityDescription,
+    NumberMode,
+)
 from homeassistant.const import PERCENTAGE
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.core import HomeAssistant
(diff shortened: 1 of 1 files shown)


fuatakgun:
That is nice, i am just worried if the user uses the up/down buttons, it would call the service for each click, right?

bouwew:
No it doesn't. I use this method in the Plugwise integration, I've tested, it waits a while until the changing is done, then the connected service is called. This is all handled in the frontend.

Wait, what is the default step-size, let me check...

bouwew:
I see you have set it to 10, no change needed, all good :)

bouwew:
My main reason for this change is the visibility of the value: the slider does not provide any info on the set value, mode BOX does.

cdnninja (approved):

bouwew:
FYI, the SOC-levels were already numbers, the default representation of a number is via a slider. The alternative is representation via a box :)

About your question, you want to avoid sending two times when a user changes both the AC and DC setpoints?
The best place to do this I think is in the `set_charge_limits()` function.
I don't have a ready-made solution for you, but I'll give it some thought.

cdnninja:
> 

Yes, we have a bug where it sends two back to back calls to the car.  The second one is rejected due to duplicate since the API is slow to process items.  I have been thinking about an easy way to bundle them.  Ideally waiting say 30 seconds and grabbing the current UI values for both to send.  I have been thinking on it but no pro just a hobby dev.  Last action tracking is one option but will take more effort still and EU doesn't support it.

PR #642, opened 2023-05-11
fix: Change odometer state_class to total_increasing

MobilettoSoft:
Since the odometer  is a value that is only destined to increase, total_increasing should be the most suitable option.

Moreover this should resolve some erratical statistics when the sensor becomes unavailable (statistics card and graphs, daily mileage when feeding an utilty meter etc)


cdnninja:
Thanks for this!

PR #669, opened 2023-07-03
chore: update dependency versions and bump version

Wolf-Code:
Updates the dependencies to use the version containing the latest fix and also bumps the version number up.


cdnninja:
Thanks for doing this.   I'm not by a pc or would.  Could you remove the version change? Our release code does that automatically.  After that I can merge it.

Wolf-Code:
Not a problem! Version change has been removed :)

PR #683, opened 2023-07-26
feat: Dutch Language Support

erikgeurts:
First iteration

### DIFF
--- a/custom_components/kia_uvo/translations/nl.json
+++ b/custom_components/kia_uvo/translations/nl.json
@@ -22,7 +22,7 @@
           "already_configured":"Het device is al geconfigureerd"
         },
        "error":{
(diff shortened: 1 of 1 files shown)


erikgeurts:
Since this is my first attempt at contributing to this repository, I'd appreciate any tips you might have to fix the failing checks.

cdnninja:
> My guess is that the problem stems from this: image

Could be!  I removed codespell since it isn't helpful for non english.  That leaves us with this as the key:

image

Any character you can swap that for to keep utf8?

cdnninja:
Quick note I have also just moved most of the services details to the translation file so this will need to be done for the dutch file too.

cdnninja:
Managed to convert to UTF 8 without loosing the character.  This seems to have resolved.  Thanks for the contribution!

cdnninja:
Do you know how we can do that using the home assistant spec?  I know it can be done for select options but not sure on others.  I am up for it if possible!

fuatakgun:
Can we refer to other strings?

https://developers.home-assistant.io/docs/internationalization/core

PR #703, opened 2023-08-13
fix: Translation for services in English

cdnninja:
Should clear up CI checks for HACS.

### DIFF
--- a/custom_components/kia_uvo/translations/en.json
+++ b/custom_components/kia_uvo/translations/en.json
@@ -26,6 +26,18 @@
           "unknown":"Unexpected error"
        }
     },
+    "services": {
+      "force_update ": {
+        "name": "Force Update",
+        "description": "Calls the car for an update to data.",
+        "fields": {
+          "device_id": {
+            "name": "Speed",
+            "description": "Target vehicle"
+          }
+        }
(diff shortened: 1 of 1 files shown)


cdnninja:
@4homeassistant @erelke @Menelao147 @toraidl    Quick note I am adding a few lines to the english translations in case you want to add to your respective work. 

Other items could be added to flush out these services but this is the minimum I need to clear some errors in the checks.

custom_components/kia_uvo/translations/en.json
```diff
@@ -26,6 +26,119 @@
           "unknown":"Unexpected error"
        }
     },
+    "services": {
+      "force_update": {
+        "name": "Force Update",
+        "description": "Force your vehicle to update its data. All vehicles on the same account as the vehicle selected will be updated.",
+        "fields": {
+          "device_id": {
+            "name": "Speed",
```
Menelao147, line 35: @cdnninja  Is "Speed" the correct value?
The other "device_id name" fields have "Vehicle" as the value.

cdnninja: Corrected.  Thanks for catching that.

PR #705, opened 2023-08-13
feat(lang): Update Italian Language

Menelao147:


cdnninja:
Quick note a few other items have been added shortly before you created this pr.  I'll merge since this takes care of most of them. 

Thanks for doing this!

PR #737, opened 2023-11-02
[Bugfix] Vehicle data sensor was only created for the last vehicle and not all

Triple-S:
Hello,
I noticed that the sensor providing the raw vehicle data from the API was only created for the last vehicle and not all of them since it was set up outside of the vehicle loop.
I'm pretty sure that this in not intended and therefore implemented a fix.
Best regards.

### DIFF
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -228,11 +228,10 @@ async def async_setup_entry(
                 entities.append(
                     HyundaiKiaConnectSensor(coordinator, description, vehicle)
                 )
+        entities.append(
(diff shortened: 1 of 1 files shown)


cdnninja:
Is this tested out locally as fixing it?

Triple-S:
I tested it in my HA installation with my car and the one of my parents and it workes fine.

Triple-S:
Is there any problem with this contribution?
I would like to sort this out as I don't want to implement the fix locally everytime I update the integration.

fuatakgun (approved):

fuatakgun:
Looks good to me, @cdnninja , any thoughts?

PR #747, opened 2023-11-12
fix: Climate heating value fix

robinlee09201:
The climate heating option -> value match should be the changed version (at least in CA). See the attached caught command. 
Screenshot 2023-11-12 170433


cdnninja:
I will merge. I can't seem to find the spot where I had found it was a 1.   Could you try the other combos to confirm descriptions match?

robinlee09201:
Just checked all the options, confirmed that "steering wheel only" is 3. Fixed in the latest commit. 
All options are checked in CA.

Issue #751, opened 2023-11-15
Possibly wrong sensor name (Monthly Energy Consumption)


Triple-S:
Please check Services, Known Bug / Issues and Troubleshooting over here first: https://github.com/fuatakgun/kia_uvo/blob/master/README.md
**Region and Brand of car**
Europe / Hyundai

**Version of the integration**
Most recent one

**Describe the bug**
I think the name of the sensor "Monthly Energy Consumption" is wrong and should actually be "Total Energy Consumption".
I have the following indications for this:

- The sensor is reading already now the field "total_power_consumed" from the Vehicle entities.
- I looked the source of this field in the hyundai_kia_connect_api up and found out that it is filled from the "responseAlltime"-request to the API which indicates that it is not just the monthly consumption which would come from the "response30d"-request to the API.
- I calculated the consumption of the most recent 30 days for my car and the one of my parents and it yields lower values than reported by the sensor "Monthly Energy Consumption" in both cases.

To avoid confucion I think the sensor should be renamed unless there is a solid reason for the naming which I don't understand yet.

PR #752, opened 2023-11-15
feat: Addition of sensor for Total Power Regenerated

Triple-S:
Hello,

this PR adds a sensor for the "total_power_regenerated" field which is already present in the API but not yet included in the Home Assistant integration. Following to the naming of the sensor for the field "total_power_consumed" which is already existing, the sensor should be named "Monthly Energy Regeneration", but as I stated in issue #751 I think there are multiple indications that the naming "Monthly" is wrong and should actually be "Total" like in the field name, so I named the sensor "Total Energy Regeneration" - Please correct me if I'm wrong.

Best regards


cdnninja:
Thank you for this!

PR #753, opened 2023-11-15
feat: Addition of sensor for the daily driving stats data.

Triple-S:
Hello,

this PR adds a sensor containing the data stored in the daily_stats field of the Vehicle objects since currently these data are not accessible in Home Assistant. The data are contained in the attributes of the sensor - one attribute per day which is identified via a datetime.date object. The values of each day are stored in a dictionay with the same keys used by the hyundai_kia_connect_api. I think having such a sensor is very useful to further process the data by e.g. templates.

Best regards

### DIFF
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -230,7 +230,9 @@ async def async_setup_entry(
                     HyundaiKiaConnectSensor(coordinator, description, vehicle)
                 )
         entities.append(
-            DailyDrivingStatsEntity(coordinator, coordinator.vehicle_manager.vehicles[vehicle_id])
+            DailyDrivingStatsEntity(
+                coordinator, coordinator.vehicle_manager.vehicles[vehicle_id]
+            )
         )
     async_add_entities(entities)
     async_add_entities(
@@ -304,6 +306,7 @@ class VehicleEntity(SensorEntity, HyundaiKiaConnectEntity):
     def unique_id(self):
         return f"{DOMAIN}-all-data-{self.vehicle.id}"
 
+
 class DailyDrivingStatsEntity(SensorEntity, HyundaiKiaConnectEntity):
     def __init__(self, coordinator, vehicle: Vehicle):
         super().__init__(coordinator, vehicle)
@@ -340,4 +343,3 @@ class DailyDrivingStatsEntity(SensorEntity, HyundaiKiaConnectEntity):
     @property
     def unit_of_measurement(self):
         return TIME_DAYS
-
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -16,6 +16,7 @@ from homeassistant.components.sensor import (
 from homeassistant.const import (
     PERCENTAGE,
     TIME_MINUTES,
+    TIME_DAYS,
     ENERGY_WATT_HOUR,
     ENERGY_KILO_WATT_HOUR,
 )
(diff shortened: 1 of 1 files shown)


cdnninja:
Thanks for this!  It has a merge conflict from another pr that just came in.

Triple-S:
@cdnninja The conflict is resolved.

cdnninja:
For some reason when I try to merge it says it can't be rebased and merged.

Triple-S:
@cdnninja I have just some basic knowledge about working with git, so I have no idea what causes this merge problem. Git tells me that there are no conflicts.
If you want I can delete this PR and the corresponding branch and recreate it from scratch from the current state of the master branch which should eliminate all merge problems.

fuatakgun:
I just squashed and merged it

gerosil:
very nice to get the daily driving stats now! thank you. How can we access the single trips to display for example as charts since they are all combined in Attributes? And how can those attributes be forwarded to influx? I know that is no UVO-specific question but maybe you had thoughts already why/how you added the daily trips as attributes

Triple-S:
@gerosil
Accessing single trip information is currently not possible. That would require changes in the underlying API (https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api) because at the moment only the daily summaries are retrieved there. The idea of this contribution was to expose data already present in the API to HA rather than adding new data.

The reason I decided to use the attributes to store the data is that this allows to easily store a variable amount of datasets. To include these data in HA in a more practical way (e.g. to display them or create statistics from them) I suggest to create template sensors to retrieve the data from the combined sensor.

The code to define a sensor for the daily consumed energy in the configuration.yaml would for example look like this:
```
template:
  - sensor:
      - name: IONIQ 5 Daily Energy Consumption
        unique_id: ioniq_5_daily_energy_consumption
        unit_of_measurement: Wh
        device_class: energy
        state_class: total_increasing
        state: >
          {% set attr = state_attr("sensor.ioniq_5_daily_driving_stats", now().date()) %}
          {{ 0 if attr == None else attr["total_consumed"]|int }}
```

gerosil:
@Triple-S 
Great! I'll give it a try. Thx!

Growiel:
Thanks for this addition, it's really interesting.

Is the `regenerated_energy` really what we gained back through regenerative breaking?

How are the numbers working relative to one another? Let's take an example:

```
2023-11-14: 
total_consumed: 10111
engine_consumption: 8758
climate_consumption: 663
onboard_electronics_consumption: 690
battery_care_consumption: 0
regenerated_energy: 10167
distance: 76
```

I have regained more energy (10167) than I have consumed (10111) over my 76km traveled that day? That can't be right, the battery % was definitely lower at the end of the day (from 90 to 75):

image

Or to the number means that I used 10111 AFTER regen braking? As in if the car did absolutely no regen braking, I would have used 20278Wh instead?

That's impressive regarless, it reduced the consumption 50%?

Triple-S:
@Growiel I think the latter is the case, but I don't know for sure. The data included in this sensor are requested from the Bluelink API and then just passed through to Home Assistant. In the end only Hyundai can tell you exactly what these data are. I noticed that in my case the consumptions listed in the Bluelink App (and since this integration is using the exact same data also here) are about 10% lower that whats displayed in my car and I have no explanation for this.

alexwhittemore:
@Growiel unfortunately, the data the API gives are suspect at best and verifiably useless at worst, so be careful relying on them for anything more than curiosity. 

For instance: distance is straight-line between the starting and ending GPS fixes. You’ll get ~0 if you drive a big loop without shutting the car off. That makes calculating I.e. efficiency from this data impossible. 

Similarly, I’ve found impossibility in samples of the other data too. For instance, I took a big trip up a mountain starting at a charger at 100% and finishing at 75%. Almost nothing was regenerated (as you’d expect driving uphill on cruise control without stopping) but the trip data said it consumed nearly 40kWh (over half the pack capacity). 

I don’t pay any attention to this data anymore, unfortunately.

PR #778, opened 2023-12-22
feat: french translation

cazzoo:
Adding the French translation file for the plugin


cdnninja:
Thank you for doing this!

PR #793, opened 2024-01-02
fix: force seats to be integers

cdnninja:
Can someone confirm this tests out? I no longer have a car to test this with.

### DIFF
--- a/custom_components/kia_uvo/services.py
+++ b/custom_components/kia_uvo/services.py
@@ -56,15 +56,15 @@ def async_setup_services(hass: HomeAssistant) -> bool:
     async def async_handle_start_climate(call):
         coordinator = _get_coordinator_from_device(hass, call)
         vehicle_id = _get_vehicle_id_from_device(hass, call)
-        duration=call.data.get("duration")
-        set_temp=call.data.get("temperature")
-        climate=call.data.get("climate")
-        heating=call.data.get("heating")
-        defrost=call.data.get("defrost")
-        front_left_seat=call.data.get("flseat")
-        front_right_seat=call.data.get("frseat")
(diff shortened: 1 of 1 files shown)


chasebolt:
@cdnninja I will try and test this out today. what is the best way to test it?

cdnninja:
Easiest way would be to edit the file in your custom_components directory with the changes here. Restart HA after that and it will take effect.

cdnninja:
I have edited the code.  Let me know if that now looks correct.  

The values supported are car dependant.

tonygauderman:
Yea, that looks correct.  I have copied the lastest file contents out and replaced my entire file and will restart and test.  For the values, I may just need to play with the numbers with a manual call.  I was testing from an automation, could test from developer tools or node red to send different values and see what happens.  Do you know how the integration determines what each vehicles valid values are?

cdnninja:
Today we have no way to know what cars support what.   We pass the value as provided.

tonygauderman:
Values are provided when the integration and vehicles are added?

cdnninja:
No values are static.  We sniffed and understood them best we could.  They could differ from car or region though.  

They are shown here.  The number is passed straight to the api and consumed by Kia Hyundai. 

https://github.com/Hyundai-Kia-Connect/kia_uvo/blob/master/custom_components/kia_uvo/services.yaml

tonygauderman:
I suspect in my case, neither of my vehicles support setting the seat settings.  They don't from the official app, and I was trying to use the integration to get around that thinking it was a limitation on the app, not my vehicles.

cdnninja:
If your native app doesn't have it the integration won't work for it either.

PR #811, opened 2024-01-31
feat(EU): Additional battery sensors

Bubo08:
In the newest pull request for the hyundai_kia_connect_api https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/489 there are three additional attributes exposed:
- SoH state of health in percentage
- Capacity as int value
- Remain (Capacity) as int value

### DIFF
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -79,6 +79,20 @@ SENSOR_DESCRIPTIONS: Final[tuple[SensorEntityDescription, ...]] = (
         device_class=SensorDeviceClass.BATTERY,
     ),
     SensorEntityDescription(
+        key="ev_battery_soh_percentage",
+        name="EV State of Health Battery",
+        native_unit_of_measurement=PERCENTAGE,
+        device_class=SensorDeviceClass.BATTERY,
+    ),
+    SensorEntityDescription(
+        key="ev_battery_remain",
+        name="EV Battery Level",
+    ),
+    SensorEntityDescription(
+        key="ev_battery_capacity",
+        name="EV Battery Capacity",
+    ),
+    SensorEntityDescription(
         key="_ev_driving_range",
         name="EV Range",
         icon="mdi:road-variant",
--- a/README.md
+++ b/README.md
@@ -39,15 +39,15 @@ These can be access by going to the developer menu followed by services.
 - set_charge_limits: You can control your charging capacity limits using this services 
 - open_charge_port / close_charge_port:  Open or close the charge port.
 
-| Service       | EU        | CA        | USA Kia        | USA Hyundai        | USA Genesis        | China |
-| ------------- | --        | --        | --------       | ------------       | -----------        |--------|
-| Update        | &#10004;  | &#10004;  | &#10004;       | &#10004;           | &#10004;           |&#10004;|
-| Force Update  | &#10004;  | &#10004;  | &#10004;       |                    |                    |&#10004;|
-| Lock Unlock   | &#10004;  | &#10004;  | &#10004;       | &#10004;           | &#10004;           |&#10004;|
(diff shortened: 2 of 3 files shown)


custom_components/kia_uvo/sensor.py
```diff
@@ -78,6 +78,20 @@
...
     ),
+    SensorEntityDescription(
+        key="ev_battery_soh_percentage",
+        name="EV State of Health Battery",
+        native_unit_of_measurement=PERCENTAGE,
+        device_class=SensorDeviceClass.BATTERY,
+    ),
+    SensorEntityDescription(
+        key="ev_battery_remain",
+        name="EV Battery Level",
```
fuatakgun, line 89: how is this different than `ev_battery_percentage`? (line 75)

Bubo08: It's a number value and I am hoping for a higher precision. My car is showing the following:
ev_battery_remain: 191700
ev_battery_capacity: 359280
ev_battery_percentage: 57.5
If I calculate then 191700 / 3592.8 = 53.35. So it seem like an interesting discrepancy. I do not know yet if this numbers are perhaps not linear and what kind of unit they are.

Bubo08: BatteryValues

custom_components/kia_uvo/sensor.py
```diff
@@ -78,6 +78,20 @@
         native_unit_of_measurement=PERCENTAGE,
         device_class=SensorDeviceClass.BATTERY,
     ),
+    SensorEntityDescription(
+        key="ev_battery_soh_percentage",
```
cdnninja, line 82: I don't see these key values in https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/master/hyundai_kia_connect_api/Vehicle.py

Without that this will fail. First it needs to be added to the vehicle, mapped in the region to that attribute.  Once done can be added here.

cdnninja (changes requested):

custom_components/kia_uvo/sensor.py
Bubo08: Now the library should be fixed. I double checked :-)

custom_components/kia_uvo/sensor.py
cdnninja: Are they helpful to expose until we know what they mean?  Do they show up in the native app anywhere?

custom_components/kia_uvo/sensor.py
cdnninja: One note so you are aware.  Now that the API is updated the manifest.json file in this repo needs the new API version.  This PR should reflect that since it is required for the sensors to function.

custom_components/kia_uvo/sensor.py
Bubo08: No they do not show up. I can only see the battery percentage in steps of .5 percentage which I think is not very precise. So I am hoping for more information. 
The SoH is also not visible in the app but I think it is a very important information.

custom_components/kia_uvo/sensor.py
Bubo08: I changed the manifest to 3.14.1.

Bubo08: Should I update the documentation? I think we might need a new column to show what is supported by the new EU version.

cdnninja: Never a bad thing.  We always are a bit behind on the docs!

Bubo08: I updated the readme. Let me know what you think.

custom_components/kia_uvo/sensor.py
cdnninja: Alright we can merge without units for now.  Maybe we will learn more with time.

cdnninja:
Huge thank you for your work @Bubo08 on this.

Issue #819, opened 2024-02-05
Add Vehicle to Load setting


jwillemsen:
Add support for Vehicle to Load as provided by the Kona 2024 EU EV. Value can be set between 20 and 80 with increments of 10, in the provided hyundai data it is the part below, full data attached. 

```
                 'Electric':
                {
                    'SmartGrid':
                    {
                        'RealTimePower': 0,
                         'VehicleToGrid':
                        {
                            'Mode': 0
                        },
                         'VehicleToLoad':
                        {
                            'DischargeLimitation':
                            {
                                'SoC': 20,
                                 'RemainTime': 0
                            }
                        }
                    }
                },
```

![IMG_2893](https://github.com/Hyundai-Kia-Connect/kia_uvo/assets/47353/8209f720-6c2f-450b-a344-157cc1d8e7d6)
[kona2024data.txt](https://github.com/Hyundai-Kia-Connect/kia_uvo/files/14162415/kona2024data.txt)

PR #821, opened 2024-02-06
feat(EU):  Vehicle2Load Discharge Limit Sensors

jwillemsen:
Expose Vehicle2Load Discharge limit, currently no set is supported, but the value is now shown correctly in HA

### DIFF
--- a/custom_components/kia_uvo/number.py
+++ b/custom_components/kia_uvo/number.py
@@ -24,6 +24,7 @@ _LOGGER = logging.getLogger(__name__)
 
 AC_CHARGING_LIMIT_KEY = "ev_charge_limits_ac"
 DC_CHARGING_LIMIT_KEY = "ev_charge_limits_dc"
+V2L_DISCHARGE_LIMIT_KEY = "ev_v2l_discharge_limit"
 
 NUMBER_DESCRIPTIONS: Final[tuple[NumberEntityDescription, ...]] = (
     NumberEntityDescription(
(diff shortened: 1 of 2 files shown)


jwillemsen:
https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/819

custom_components/kia_uvo/number.py
```diff
@@ -103,14 +113,22 @@ async def async_set_native_value(self, value: float) -> None:
...
 
         # set new limits
         if self._description.key == AC_CHARGING_LIMIT_KEY:
             ac = value
             dc = self.vehicle.ev_charge_limits_dc
-        else:
+        if self._description.key == DC_CHARGING_LIMIT_KEY:
             ac = self.vehicle.ev_charge_limits_ac
             dc = value
+        if self._description.key == V2L_DISCHARGE_LIMIT_KEY:
```
cdnninja, line 129: If the above now filters to the key does this need to exist? Haven't looked closely at the code though.  

My impression is this if statement doesn't do anything.

jwillemsen: Code above guards setting the same value, in this place later we can add the code to set the value to the lower layer

Bubo08: I think the code is correct and could be merged. 

I would prefere the use of elif which makes it more readable but in the code above it is also not used.

fuatakgun:
> Expose Vehicle2Load Discharge limit, currently no set is supported, but the value is now shown correctly in HA

if set is not supported, let's expose it as SENSOR rather than NUMBER

jwillemsen:
> > Expose Vehicle2Load Discharge limit, currently no set is supported, but the value is now shown correctly in HA
> 
> if set is not supported, let's expose it as SENSOR rather than NUMBER

Hopefully we can make it supported soon (currently not having enough time), making it now a number and change it in some weeks to a sensor leads to an unavailable sensor at that moment.

fuatakgun:
I don't think it is a big thing to worry about?

jwillemsen:
Ok, will create a new branch with that idea asap

Issue #824, opened 2024-02-09
Include state_class: measurement into sensor.XXXXX_ev_battery_level entity


salfuman33:
**Is your feature request related to a problem? Please describe.**
I tried to add this entity to a new card into Home Assistant as I do with my other Cars or devices, but it doesnt works because there are no statistics about it.

**Describe the solution you'd like**
I have solve it modifiying the entity manualy and adding  state_class: measurement attribute to it

**Describe alternatives you've considered**
Probably the entity should have this attribute by default as the other battery level entities that I have by other devices and integrations.

**Additional context**

cdnninja:
Are you able to create a pull request for this?  I believe quick change.

salfuman33:
> Are you able to create a pull request for this? I believe quick change.

Sorry im not fluent in Github. I think i cant do it and yes I think its a quick change and necessary to report as other battery identities from other vendors.

jwillemsen:
I think this one can be closed

PR #831, opened 2024-02-16
feat: Add Measurement state_class to EV Battery Level

gzumba:
We want to have state_class defined to have it stored in long term statistics.

More information:
https://developers.home-assistant.io/blog/2021/05/25/sensor_attributes/


gzumba:
This fixes Issue #824

PR #857, opened 2024-03-30
feat: add start/stop charge and charge limits to Hyundai USA

stephenjamieson:


stephenjamieson:
Hrm, maybe I shouldn't have bumped the version. Up to you.

cdnninja:
Correct just do the api.  Version of the integration is managed by a ci in GitHub.

stephenjamieson:
Reverted that version change, I did change the readme so that it reflects the change, however it looks like my editor replaced the html code with the actual symbols, so I can remove the readme change if you like.

PR #895, opened 2024-06-20
feat: calls update state on completion

rjantoine:
Calls to lock, unlock, start_climate, stop_climate, start_charge, stop_charge and set_charge_limits now update the state of the vehicle when complete.

TL/DR: ``await async_request_refresh()`` has been replaced by ``don't await, async try to wait until action is done and finally async_refresh()``. The try block should ensure that this feature is completely neutral or additive.

This was done by adding a new ``async_await_action_and_refresh`` function to the coordinator. It adds ``check_action_status`` that awaits for success or error before calling ``async_refresh``. This function is called using ``async_create_task`` so that the action can finish without having to wait for a response that can take about 30 to 60 seconds. Since the function now waits for the action to complete before refreshing, the refreshed state now has the updated (post-action) information, updating the UI with the new status. To help avoid any possible bugs in regions that don't have check_action_status properly implemented ([HyundaiBlueLinkAPIUSA](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/master/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py) which doesn't have that function and [KiaUvoAPIUSA](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/master/hyundai_kia_connect_api/KiaUvoAPIUSA.py) which doesn't implement the synchronous feature), check_action_status is wrapped in a try block, with a finally block that calls for a refresh, just as it did before.

This has been tested with lock, unlock, start_climate and stop_climate in CA and works as expected.

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -167,63 +167,69 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
             self.vehicle_manager.check_and_refresh_token
         )
 
+    async def async_await_action_and_refresh(self, vehicle_id, action_id):
+        await self.hass.async_add_executor_job(
+            self.vehicle_manager.check_action_status, vehicle_id, action_id, True, 60
+        )
+        await self.async_refresh()
+
     async def async_lock_vehicle(self, vehicle_id: str):
         await self.async_check_and_refresh_token()
-        await self.hass.async_add_executor_job(self.vehicle_manager.lock, vehicle_id)
-        await self.async_request_refresh()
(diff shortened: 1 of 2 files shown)


rjantoine:
Now we just need to change version number. 😃 Not sure how to do that though.

cdnninja:
It does that Sundays and Wednesdays automatically.  I can force it too but the delay helps others test as they can use the have service call and update from "master".

cdnninja:
As a note for awareness. The system looks at pr titles for version number and release notes.  Feat: triggers a 2.x release.  Fix: triggers a 2.5.x release.    Chore: docs: and a few others don't cause a release. 

Release notes are created based on pr titles.

Issue #864, opened 2024-04-12
Change AC-charge-current


dagobert:
It is possible to change am the AC charge current to 100%, 90% and 60% when sitting in the car using the settings in EV screen on an Ioniq 5 and others with the same platform.

This option would've great to expose to HA to be able to regulate the charge current when loading at home with solar. 

Is it in the API? How to find out?

dagobert:
I am not sure I understand you correctly but there is a way to do what I wrote from the car screen. Just follow the screenshots and you will find it. Sorry the ui is in German.

From ev screen use the burger icon.
![IMG_20240419_162852](https://github.com/Hyundai-Kia-Connect/kia_uvo/assets/140021/d07e78d8-993d-426f-b5d1-1ee075219416)

There you use ev-settings.
![IMG_20240419_162801](https://github.com/Hyundai-Kia-Connect/kia_uvo/assets/140021/9404bf65-92a8-49e9-8999-ca5369f057b5)

Than ac-current
![IMG_20240419_162715](https://github.com/Hyundai-Kia-Connect/kia_uvo/assets/140021/5e48ddde-ac0a-4a93-ac91-fdfd583c0984)

Here you can change the percentage of the max available current of the charging process. This actually means that it uses x percent of what the charger offers at this right moment and not of what it could offer at maximum. Hence, if you setup your simple plugin charger to 8 amps at 230v than it would use this as 100 percent and the lower accordingly. This paragraph has nothing to do with my request but is for you to understand the mechanism of these settings options.
![IMG_20240419_162632](https://github.com/Hyundai-Kia-Connect/kia_uvo/assets/140021/26d557ea-e09e-45dc-88d6-6d4a57f7f90d)

Speederer:
Okay, you are right, I skipped over that setting. Sorry for that.
It looks promising for controlling the charging, but too bad it's not supported remotely.

cdnninja:
I assume that is different than the ac /dc charge setting we offer in this today?

dagobert:
Yes, the difference is, that the charge option sets the maximum charge point to whî the car should be filled, eg. 80% full.

The option I talk about sets the current. It basically the speed the car is charging (in this case related to the amps the charger offers).

cdnninja:
Someone will need to sniff this traffic and add it.

cdnninja:
try updating to master branch.  That was an error in the code.

ReconOperator:
@cdnninja 

[...]

PR #896, opened 2024-06-24
feat(EU): Support setting the charging current

jrnk:
Following up with this to support https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/commit/632951ff7a970541dec7bcde0960c47bcbaa71a2 

Should solve https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/864

Note: I've translated all but the Chinese (that was missing more translation keys), but the indentation seems to be different for every localization which was a bit cumbersome, but I didn't want to clutter this commit with formatting.


cdnninja:
Looks good.  Would it be worth adding a sensor to display it? Either a number slider so you can edit or just a sensor to display.  

Number with edit may be hard since only a few values work, I guess drop down menu could be an option too.

jrnk:
Yes, good idea, will add it!

cdnninja:
Same pr or different?

cdnninja:
Also the manifest file needs to be updated to the new api version.

jrnk:
> Also the manifest file needs to be updated to the new api version.

All will be in this PR, already bumped the version in the manifest

jrnk:
@cdnninja added the sensor, also needed to add a mapping function for it, if you have another suggestion to do this: please shoot, this is new for me ;)

jrnk:
> Number with edit may be hard since only a few values work, I guess drop down menu could be an option too.

The selector was already added in `services.yml`, does it need more for it to work?

cdnninja:
> > Number with edit may be hard since only a few values work, I guess drop down menu could be an option too.
> 
> 
> 
> The selector was already added in `services.yaml`, does it need more for it to work?
> 
> 
> 
> 


Services work but people often want it on the dashboard.  We can do without though, you have done the heavy lift.

cdnninja:
> > Number with edit may be hard since only a few values work, I guess drop down menu could be an option too.
> 
> 
> 
> The selector was already added in `services.yaml`, does it need more for it to work?
> 
> 
> 
> 

Mapping is often done as enum in the constants file in both the api and home assistant files.

custom_components/kia_uvo/sensor.py
```diff
@@ -240,6 +240,13 @@
         native_unit_of_measurement=PERCENTAGE,
         device_class=SensorDeviceClass.BATTERY,
     ),
+    SensorEntityDescription(
+        key="ev_charging_current",
+        name="Charging current limit",
```
cdnninja, line 245: We use capitals for each letter for exposed names.

custom_components/kia_uvo/sensor.py
```diff
@@ -372,3 +379,8 @@ def unique_id(self):
     @property
     def unit_of_measurement(self):
         return UnitOfTime.DAYS
+
+
+def charging_current_mapper(value):
+    mapping = {1: 100, 2: 90, 3: 60}
```
cdnninja, line 385: This should move to constants.

custom_components/kia_uvo/services.yaml
```diff
@@ -226,3 +226,21 @@ set_charge_limits:
...
+      selector:
+        device:
+          integration: kia_uvo
+    level:
+      required: true
+      selector:
+        select:
+          options:
+            - label: "100%"
+              value: 1
```
cdnninja, line 242: Not sure why but the CI isn't happy with this: https://github.com/Hyundai-Kia-Connect/kia_uvo/actions/runs/9652140110/job/26621566236

cdnninja (changes requested):
Here is a bit of feedback. It is very close! You have learned quick.   Help on this integration is appreciated this I no longer have one of these cars.

jrnk:
@tobixx yes saw that! Fixed in https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/581

jrnk:
@tobixx ah I missed you could not set it in the app, then it's not available for the EV6 I guess. For an EV9 you can control this current in the app on the "EV service" page. 

image

Issue #870, opened 2024-04-30
Daily Energy consumption


cazzoo:
I'm tracking my both kia EV & PHEV cars and I wonder if we could implement a daily energy consumption (both KWH and fuel).
Indeed, I'm trying to add my EV consumption in HA Energy tab but right now the only sensor is a mean value of the total KWH/km.
This is not relevant.

**Describe the solution you'd like**
Have a sensor tracking daily use of the electricity used by the car.

**Include the logs of the end point from Data in HA if this is a sensor**
?

ZuinigeRijder:
You can use [hyundai_kia_connect_monitor](https://github.com/ZuinigeRijder/hyundai_kia_connect_monitor?tab=readme-ov-file#introduction-hyundai_kia_connect_monitor).

cazzoo:
Cool but seems overkill for what I was looking to do. It seems there's no better way to handle it. Thanks for the link anyway

Triple-S:
The data should already be provided by the integration via the daily driving stats sensor - It might be not available for specific regions, but for EU it is. In that case you just need a sensor extracting the consumtion of the current day from the attributes of the driving stats sensor which can be achieved by a template sensor:
```
{% set attr = state_attr("sensor.[vehicle]_daily_driving_stats", now().strftime("%Y-%m-%d")) %}
{{ attr["total_consumed"]|int(0) if attr != None else 0 }}
```

cazzoo:
That's an interesting sensor to put in place, let me try. Thanks for the hint

PR #954, opened 2024-10-27
feat: Add Sensor for todays daily stats

MichaelMKKelly:
Adds a sensor for "Todays Daily Driving Stats" which allows for a fixed sensor to access daily stats instead of the moving target of the date labeled attribute in the main daily driving stats sensor.

This makes it easier to pull the current days stats and have it reset at midnight as it will switch to default 0 values. instead of using template workarounds such as those discussed in #870 or other helper implementations people may be using.

the entity gets added along with the normal Daily Driving Stats so it only exists if that does and draws data using the same method.

NB: 1st PR on a HA integration, please be kind.

### DIFF
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -364,8 +364,6 @@ class DailyDrivingStatsEntity(SensorEntity, HyundaiKiaConnectEntity):
         m = {}
         for day in self.vehicle.daily_stats:
             key = day.date.strftime("%Y-%m-%d")
-            today = date.today()
-            todayskey = today.strftime("%Y-%m-%d")
             value = {
                 "total_consumed": day.total_consumed,
                 "engine_consumption": day.engine_consumption,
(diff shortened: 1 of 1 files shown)


cdnninja (changes requested):
I believe this sensor will be added and set to 0 for cars that don't have driving stats.  Could you add a if statement to check for stats before appending?

cdnninja:
Actually scratch that.  I see an existing if is used.  

Let's merge and see what happens! I can't test since I don't have a Kia anymore.

Triple-S:
What is the reason for adding a new sensor that just copys already existing data?
I really don't see the point of this change besides blowing up the database unnecessarily.

MichaelMKKelly:
> What is the reason for adding a new sensor that just copys already existing data? I really don't see the point of this change besides blowing up the database unnecessarily.

so there is a single target to get "todays data" rather then go looking for the correct attribute of a sensor which may or may  not exist. if there is no data for "today" this returns 0. having a single target that resets at midnight simplifies obtaining and tracking daily data.

1 sensor with a few attributes is hardly "blowing up the database" . You can disable the sensor if you don't need/want it. although I suppose there is maybe an argument for disabling it by default and people switching it on if they want it. whichever way around it is there will likely be disagreement.

PR #960, opened 2024-11-03
fix: usa climate start

renatkh:


cdnninja:
Oops.  I should have looked closer.  No need to bump this version number.   The release process does that.   

When by a pc I'll try force a release so HACS recognizes it.

PR #1010, opened 2025-01-12
feat: Adds steering wheel heat as its own selector

erikn13:
to allow for Low/High setting. Related PR for hyundai_kia_connect_api/feature/steering_wheel_step

### DIFF
--- a/custom_components/kia_uvo/services.yaml
+++ b/custom_components/kia_uvo/services.yaml
@@ -59,10 +59,12 @@ start_climate:
           options:
             - label: "Off"
               value: "0"
-            - label: "Side and Back Defroster"
+            - label: "Steering Wheel(noneEU), Side and Back Defroster"
               value: "4"
             - label: "Rear Window Only"
               value: "2"
+            - label: "Steering Wheel Only(None EU, for US use steering wheel option)"
+              value: "3"
     steering_wheel:
       required: false
       example: 1
--- a/custom_components/kia_uvo/services.yaml
+++ b/custom_components/kia_uvo/services.yaml
@@ -63,7 +63,7 @@ start_climate:
               value: "4"
             - label: "Rear Window Only"
               value: "2"
-    steeringWheel:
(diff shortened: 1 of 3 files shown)


custom_components/kia_uvo/services.yaml
```diff
@@ -59,12 +59,23 @@ start_climate:
           options:
             - label: "Off"
               value: "0"
-            - label: "Steering Wheel, Side and Back Defroster"
```
cdnninja, line 62: I think this impacts other regions.  I'm on my phone so hard to check but could you validate?

erikn13:
start_climate is associated with EU, CA, USA Kia, USA Hyundai, and China according to the readme

custom_components/kia_uvo/services.yaml
erikn13: Upon further investigation, these particular features only exist in this form, in the KiaUvoAPIUSA

cdnninja: So this would impact USA users?

erikn13: Yes. I didn't see anything region specific in `kia_uvo` custom_component. The only region edited in the `hyundai_kia_connect_api` was the USA but the `ApiImpl.py` was also modified to add the option for `steeringWheel`. The `options.heating` remains the same in other territories, this just moves the steering wheel option out of that grouping to allow for granularity. I didn't see any calls related to the steering in the other regions in the `hyundai_kia_connect_api`. I couldn't think of a clean backwards compatability for the `steeringWheel` portion of the body in `KiaUvoApiUSA.py`.

cdnninja:
Clarifying these changes in some way are only EU is important.   As well making the strings clear.

cdnninja:
Adjusted this language to show what is EU and what isn't.    Merging to solve an issue for someone but open to further refinements on this.  Doc updates would also be good.

Issue #1029, opened 2025-02-03
500 Internal Server when trying to configure sacn interval


lsmllc:
**Region and Brand of car**
CA / Hyundai Tucson
**Version of the integration**
2.35.1
**Describe the bug**
When trying to configure scan interval I get the following error : Config flow could not be loaded: 500 Internal Server Error Server got itself in trouble
**Debug logs if an error occurred**
Logger: aiohttp.server
Source: /usr/local/lib/python3.13/site-packages/aiohttp/web_protocol.py:451
First occurred: 12:02:10 PM (9 occurrences)
Last logged: 1:57:11 PM

Error handling request
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_protocol.py", line 480, in _handle_request
    resp = await request_handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_app.py", line 569, in _handle
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/http/security_filter.py", line 92, in security_filter_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/http/forwarded.py", line 83, in forwarded_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/http/request_context.py", line 26, in request_context_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/http/ban.py", line 86, in ban_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/http/auth.py", line 242, in auth_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/http/headers.py", line 32, in headers_middleware
    response = await handler(request)
[...]

PR #1030, opened 2025-02-04
fix: non-working config flow

blind-oracle:
Make it consistent with current Home Assistant API & practices.

Fixes https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1029

I didn't bump the version, probably you'll do you own release process.

Tested to work fine:
image


cdnninja:
Correct on version.   GitHub actions will do that twice a week automatically.

cdnninja:
Thanks for the work on this!

PR #1083, opened 2025-04-08
feat(EU): V2L

cdnninja:
Add V2L as a number sensor


cdnninja:
@jwillemsen  could you test this out?

jwillemsen:
Sorry, but unable to test this, I disabled this integration as it pulled the 12V battery down to zero twice during the last months and I had to call someone to jump start the Kona

PR #1093, opened 2025-04-17
fix: V2l for CCS vehicles

rutj87:
update services.py and services.yaml

### DIFF
--- a/custom_components/kia_uvo/services.py
+++ b/custom_components/kia_uvo/services.py
@@ -268,6 +268,7 @@ def async_setup_services(hass: HomeAssistant) -> bool:
         coordinator = _get_coordinator_from_device(hass, call)
         vehicle_id = _get_vehicle_id_from_device(hass, call)
         await coordinator.async_set_vehicle_to_load_discharge_limit(vehicle_id)
+        v2l_discharge_limit = call.data.get("v2L_limit")
 
     services = {
         SERVICE_FORCE_UPDATE: async_handle_force_update,
--- a/custom_components/kia_uvo/services.py
+++ b/custom_components/kia_uvo/services.py
@@ -287,7 +287,7 @@ def async_setup_services(hass: HomeAssistant) -> bool:
         SERVICE_START_HAZARD_LIGHTS_AND_HORN: async_handle_start_hazard_lights_and_horn,
         SERVICE_START_VALET_MODE: async_handle_start_valet_mode,
         SERVICE_STOP_VALET_MODE: async_handle_stop_valet_mode,
-        SERVICE_SET_VEHICLE_TO_LOAD_DISCHARGE_LIMIT_MODE: async_handle_set_vehicle_to_load_discharge_limit,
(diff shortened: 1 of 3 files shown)


rutj87:
@cdnninja make sure you check this!

Also, I may be missing a call.get.data entry around line 271?

cdnninja:
This wouldn't be required but rather a new feature.   It was set so you would use the set number or whatever it is called as an action call.    If you want a distinct service for it would need this though.  I only created the basics for V2L to get the number slider going.   

The bug for number slider was the name didn't actually match the method in coordinator file.    I also found a naming consistency issue in the set charge limits method which is why I made the mistake.  I am fixing that here: https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/1094

This PR could change to a "feat:" to add the service though if you would like.

custom_components/kia_uvo/services.py
```diff
@@ -262,6 +264,12 @@ async def async_handle_stop_valet_mode(call):
         vehicle_id = _get_vehicle_id_from_device(hass, call)
         await coordinator.async_stop_valet_mode(vehicle_id)
 
+    async def async_handle_set_vehicle_to_load_discharge_limit(call):
+        coordinator = _get_coordinator_from_device(hass, call)
+        vehicle_id = _get_vehicle_id_from_device(hass, call)
+        await coordinator.async_set_vehicle_to_load_discharge_limit(vehicle_id)
+        v2l_discharge_limit = call.data.get("v2l_limit")
```
cdnninja, line 271: This line would move above the action of the coordinator.  As per this line that method expects vehicle_id, followed by the value.  Value being v2l_discharge_limit.

https://github.com/Hyundai-Kia-Connect/kia_uvo/blob/20c7f3233347278e7fdc9b18c71960373661c3fe/custom_components/kia_uvo/coordinator.py#L330

custom_components/kia_uvo/services.yaml
```diff
@@ -410,3 +410,18 @@ stop_valet_mode:
...
           integration: kia_uvo
+set_vehicle_to_load_discharge_limit:
+  fields:
+    device_id:
+      required: false
+      selector:
+        device:
+          integration: kia_uvo
+    v2l_limit:
+      required: fale
```
cdnninja, line 421: Should be required.  Otherwise it will fail.  Should be true.

custom_components/kia_uvo/translations/en.json
```diff
@@ -317,8 +317,20 @@
         "device_id": {
           "name": "Vehicle",
           "description": "Target vehicle"
+    },
+    "set_vehicle_to_load_discharge_limit": {
+      "name": "Set V2L Discharge Limit",
+      "description": "sets V2L disharge limit",
```
cdnninja, line 323: discharge

cdnninja (changes requested):

rutj87:
Ok I'll close this one

PR #1109, opened 2025-04-21
feat!: Switch select sensors to diagnostic category

RustyDust:
Mostly influenced by personal taste, so feel free to keep what you like and trash the rest ;)
The main reason for the changes below is to be able to easier distinguish between cars, especially if there's more than one of the same model.

## Changes

- Add VIN to sensor list
- Add the (nick)name to the displayed device info for easier distinction in case one owns two cara of the same model
- Add the VIN as serial number to the device info

| Before | After |
| ------- | ----- |
| old_001 | new_001 |

- Move most of the binary sensors to the diagnostic section of the entities

| Before | After |
| ------- | ----- |
| old_002 | new_002 |

> [!NOTE]
> Although the internal names of the sensors now to be found under "Diagnostics" won't be changed by the new _entity_class_ the internal statistics for those may be reset!

### DIFF
--- a/custom_components/kia_uvo/entity.py
+++ b/custom_components/kia_uvo/entity.py
@@ -21,5 +21,6 @@ class HyundaiKiaConnectEntity(CoordinatorEntity):
             identifiers={(DOMAIN, self.vehicle.id)},
             manufacturer=f"{BRANDS[self.coordinator.vehicle_manager.brand]} {REGIONS[self.coordinator.vehicle_manager.region]}",
             model=f"{self.vehicle.name} ({self.vehicle.model})",
+            name=f"{self.vehicle.name} ({self.vehicle.model})",
             serial_number=f"{self.vehicle.VIN}",
         )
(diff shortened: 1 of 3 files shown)


custom_components/kia_uvo/entity.py
```diff
@@ -20,6 +20,6 @@ def device_info(self):
         return DeviceInfo(
             identifiers={(DOMAIN, self.vehicle.id)},
             manufacturer=f"{BRANDS[self.coordinator.vehicle_manager.brand]} {REGIONS[self.coordinator.vehicle_manager.region]}",
-            model=self.vehicle.model,
-            name=self.vehicle.name,
```
cdnninja, line 24: Why would name get removed?

cdnninja (changes requested):
Overall looks good.   I question moving name away from the official spot for it into model.

custom_components/kia_uvo/entity.py
RustyDust: Oops. Forgot to push my latest changes after experimenting with the layout. Fixed, thanks!

cdnninja:
Are you able to test if the data gets reset with this?  If it does I think we either need migration or list as breaking change. I would think if unique name is unchanged it should carry over.

RustyDust:
Sorry for the delay. To answer your question: I tested on three different installations and the values are carried over without problem. However, the storage period for historical data seems to get limited to a week. Considering that all the "diagnostic" values are simple "On/Off" or "Open/Closed" states I don't think that's a problem since I can't imagine a use case where I'd like to see for example how often the car doors were opened in a day.
Still, this could be considered a breaking change.

PR #1187, opened 2025-07-07
chore: upgrade library and bump version

johanfleury:
This should fix login issues in Canada.


cdnninja:
Pypi uploaded failed for some reason so this won't work either. 


As a note the GitHub action updates the version of the integration as it releases in GitHub.

Issue #1213, opened 2025-08-01
kia_uvo version 2.43.1 causes errors


tbclark3:
Please check Services, Known Bug / Issues and Troubleshooting over here first: https://github.com/fuatakgun/kia_uvo/blob/master/README.md
**Region and Brand of car** USA/Hyundai

**Version of the integration**  2.43.1

**Describe the bug**
Following upgrade to 2.43.1, the integration consistently causes errors on startup. This is apparently a new requirement, so I cannot compare to the previous version. Reverting to the previous version causes the error to go away.

**Debug logs if an error occurred**
2025-07-31 15:22:10.813 ERROR (SyncWorker_7) [homeassistant.util.package] Invalid requirement 'git+https://github.com/VeNoMouS/cloudscraper.git@3.0.0'
2025-07-31 15:22:28.933 ERROR (SyncWorker_9) [homeassistant.util.package] Invalid requirement 'git+https://github.com/VeNoMouS/cloudscraper.git@3.0.0'
2025-07-31 16:00:00.387 ERROR (MainThre

Outlined here: https://github.com/fuatakgun/kia_uvo/blob/master/README.md

**To Reproduce**
Upgrade to 2.43.1 and restart HA.

**Expected behavior**
HA should start without errors.

**Screenshots**

**Additional context**

cdnninja:
What version of home assistant are you running and what are you hosting it on?

distanzcheck:
Dieser Fehler stammt von einer benutzerdefinierten Integration

Logger: custom_components.kia_uvo.coordinator
Quelle: helpers/update_coordinator.py:382
Integration: Hyundai / Kia Connect (Dokumentation, Probleme)
Erstmals aufgetreten: 08:42:05 (121 Vorkommnisse)
Zuletzt protokolliert: 11:09:05

Unexpected error fetching kia_uvo data
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 183, in login
    authorization_code = self._get_authorization_code_with_redirect_url(
        username, password, cookies
    )
  File "/usr/local/lib/python3.13/site-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 1102, in _get_authorization_code_with_redirect_url
    parsed_url = urlparse(response["redirectUrl"])
                          ~~~~~~~~^^^^^^^^^^^^^^^
[...]

PR #1236, opened 2025-08-08
fix: Canada Login

alphalneo:
Fix the manifest to load correctly the cloudscraper library in HAOS

### DIFF
--- a/custom_components/kia_uvo/manifest.json
+++ b/custom_components/kia_uvo/manifest.json
@@ -8,6 +8,6 @@
   "iot_class": "cloud_polling",
   "issue_tracker": "https://github.com/Hyundai-Kia-Connect/kia_uvo/issues",
   "loggers": ["kia_uvo", "hyundai_kia_connect_api"],
-  "requirements": ["hyundai_kia_connect_api==3.43.5"],
+  "requirements": ["hyundai_kia_connect_api==3.43.6"],
   "version": "2.43.4"
 }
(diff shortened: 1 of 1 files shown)


cdnninja:
With this is breaks for some, my impression docker based installs?

https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1213

Maybe we take the lesser of two issues based on reports.

alphalneo:
Sorry I don't know how to solve it, I tried the docker version, and this method is working, when I do a pip3 freeze I see this
cloudscraper @ git+https://github.com/VeNoMouS/cloudscraper.git@9ea528a8675f1bebd49ff853d142e94988a95178

Edited: I found this https://github.com/home-assistant/core/issues/84121. What do you think about not using cloudscraper, but instead using requests, modifying the headers, and including the user agent? Or using the old one?

alphalneo:
Another option is to use cloudscraper V1 at the moment I am able to authenticate, if you consider an option I will make some test

cdnninja:
Very open to it if you can get it working.  Since I don't have a car or this issue hard for me to tinker to see combos that work.  Ideally we don't use cloudscraper.

alphalneo:
I have it running with the old version of Cloudscraper. I modified the manifest and added a User-Agent in KiaUvoApiCA.py), since the one generated doesn’t pass Cloudflare (Cloudscraper v1 generates a different User-Agent, but it fails for some reason).

cdnninja:
Could you create a pull request for the user agent change? Then we can remove this requirement from this manifest file?

PR #1275, opened 2025-08-25
fix: Send integers instead of strings to control the windows (Kia EV6/Germany)

marvinwankersteen:
Kia's API expects integers instead of strings to control the windows. I checked that with captured traffic of the App "Kia Connect".
Here, only the type is changed via the GUI. You can also set integers using the YAML editor, and then it will work.

I could imagine that this does not apply to Hyundai, for example. However, I cannot verify this.

See https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/852 for more details.

### DIFF
--- a/custom_components/kia_uvo/manifest.json
+++ b/custom_components/kia_uvo/manifest.json
@@ -9,5 +9,5 @@
   "issue_tracker": "https://github.com/Hyundai-Kia-Connect/kia_uvo/issues",
   "loggers": ["kia_uvo", "hyundai_kia_connect_api"],
   "requirements": ["hyundai_kia_connect_api==3.45.1"],
-  "version": "2.44.1"
+  "version": "2.45.0"
 }


fuatakgun:
according to definition, values are supposed to be string (https://www.home-assistant.io/docs/blueprint/selectors#value).

I suggest, reverting back to old service.yaml and parsing the integers inside services.py.

While doing so, I also suggest renaming variables into the convention.
```
flseat -> front_left_seat
frseat -> front_right_seat 
rlseat -> rear_left_seat 
rrseat -> rear_right_seat 
```

lastly, thanks a lot for your contributions 👏 👏 👏 👏 👏

marvinwankersteen:
Hmm, I don't understand or can't figure out why GitHub tried to do a merge. I only synced my fork with this repo. Was that probably the reason?

It's good that the validation failed here.

PR #1292, opened 2025-09-02
fix(EU): Store Refresh Token

cdnninja:


cdnninja:
Other option with this is build a feature to set within the API and store it under the correct name. In theory that would reduce logins for all regions.

marvinwankersteen:
@cdnninja:
The if statement does not work here because `BRAND_KIA` and `REGION_EUROPE` is set incorrectly (or `brand` and `region`).

My changes
```diff
async def async_unload_entry(hass: HomeAssistant, config_entry: ConfigEntry) -> bool:
    """Unload a config entry."""

    """If config entry is Kia EU, update password field with current refresh token."""
+   _LOGGER.debug(
+       f"{DOMAIN} - async_unload_entry executed."
+   )
    manager = hass.data[DOMAIN][config_entry.unique_id].vehicle_manager
    brand = config_entry.data.get(CONF_BRAND, "")
    region = config_entry.data.get(CONF_REGION, "")
+   _LOGGER.debug(f"{DOMAIN} - Brand: {brand}/{BRAND_KIA}")
+   _LOGGER.debug(f"{DOMAIN} - Region: {region}/{REGION_EUROPE}")
+   if brand == 1 and region == 1:
-   if brand == BRAND_KIA and region == REGION_EUROPE:
        refresh_token = manager.token.refresh_token
        new_data = dict(config_entry.data)
        new_data[CONF_PASSWORD] = refresh_token
+       _LOGGER.debug(f"{DOMAIN} - Refresh Token: {refresh_token}")
        hass.config_entries.async_update_entry(config_entry, data=new_data)
        _LOGGER.debug(
            f"{DOMAIN} - Updated password field with refresh token for Kia EU config entry."
        )

    if unload_ok := await hass.config_entries.async_unload_platforms(
        config_entry, PLATFORMS
    ):
        del hass.data[DOMAIN][config_entry.unique_id]
    if not hass.data[DOMAIN]:
        async_unload_services(hass)

    return unload_ok
```
result in the following output:
```
2025-09-02 09:51:32.959 DEBUG (MainThread) [custom_components.kia_uvo] kia_uvo - async_unload_entry executed.
2025-09-02 09:51:32.959 DEBUG (MainThread) [custom_components.kia_uvo] kia_uvo - Brand: 1/Kia
2025-09-02 09:51:32.959 DEBUG (MainThread) [custom_components.kia_uvo] kia_uvo - Region: 1/Europe
2025-09-02 09:51:32.959 DEBUG (MainThread) [custom_components.kia_uvo] kia_uvo - Refresh Token: <REFRESH_TOKEN>
2025-09-02 09:51:32.959 DEBUG (MainThread) [custom_components.kia_uvo] kia_uvo - Updated password field with refresh token for Kia EU config entry.
```

I can't say whether the password was actually saved, because I already have to enter the refresh_token in the password field, otherwise the integration won't load.
But `manager.token.refresh_token` is set with the correct refresh_token that I entered in the password field.

fuatakgun:
I would do nothing inside kia_uvo, other than informing users on the issue and redirecting them to a specific wiki page under github repository to explain them how to retrieve refresh token.

They should enter the extracted the refresh token into password field and the rest would work.

cdnninja:
I will put this as draft for now.    My impression previously was the refresh token would change when we run check and refresh token. However that method may always get the same token forever.  Some system leave the refresh forever, others do expire it and issue new ones. 

If we get reports of it expiring at some point when a reboot happens we will know this is needed.  Until that time not worth merging.

fuatakgun:
> I will put this as draft for now. My impression previously was the refresh token would change when we run check and refresh token. However that method may always get the same token forever. Some system leave the refresh forever, others do expire it and issue new ones.
> 
> If we get reports of it expiring at some point when a reboot happens we will know this is needed. Until that time not worth merging.

as I said, if I had implemented a proper tokenization mechanism, refresh token should not expire at all and we should always get a new short living access token.

I was too lazy at the time and called the login flow when access token was expired, instead of generating a new accees token using the refresh token.

Overall oauth flow;
1. from home assistant to kia, call it with a specific redirect URL
2. user logs in kia website and kia redirects the user to initial redirect URL and append a code
3. we should use the code to get refrech token and access token
4. refresh token almost equal to username,password. it must be stored with extra safety in mind.
5. access token is short lived and should be renewed using refresh token.

PR #1310, opened 2025-09-17
fix: Update climate start options to allow longer duration

kermitageo:
The current version of the US MyHyundai app allows climate start duration to be up to 30 minutes, in 5 minute intervals. The HACS integration currently supports a maximum of 10 minutes. 

IMG_0492

### DIFF
--- a/custom_components/kia_uvo/services.yaml
+++ b/custom_components/kia_uvo/services.yaml
@@ -25,9 +25,9 @@ start_climate:
       default: 5
       selector:
         number:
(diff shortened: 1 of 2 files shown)


cdnninja:
I don't think other regions are the same.   The initial settings were based on other regions.  Could we increase the max but not change the other values for now?

kermitageo:
Yes, I think that's fine. Just increased the max duration and added more description to the field.

PR #1346, opened 2025-10-15
fix: bump hyundai_kia_connect_api to 3.48.0

Maaxion:
Update hyundai_kia_connect_api to latest version to fix login in EU

Once this is merged, refresh_token can be used in place of password to authenticate.

This requires users to first fetch their refresh_token.

### DIFF
--- a/custom_components/kia_uvo/requirements.txt
+++ b/custom_components/kia_uvo/requirements.txt
@@ -1,1 +1,1 @@
-hyundai_kia_connect_api==3.44.5
+hyundai_kia_connect_api==3.48.0


fuatakgun:
Thanks again for your support, I went ahead just a couple of minutes ago and bumped the version already.

dMopp:
Please bump to 3.48.1, because .0 killed KIA Login again :(

custom_components/kia_uvo/requirements.txt
```diff
@@ -1 +1 @@
-hyundai_kia_connect_api==3.44.5
+hyundai_kia_connect_api==3.48.0
```
mathieupotier:
```suggestion
hyundai_kia_connect_api==3.48.1
```

fuatakgun: while we have a requirements.txt to be consumed, we did not set this library properly, so, it is coming from this file for home assistant: https://github.com/Hyundai-Kia-Connect/kia_uvo/blob/master/custom_components/kia_uvo/manifest.json

Ros-83:
Anche effettuando l'aggiornamento non funziona.Ho controllato le api è sono aggiornate per come segnalato nel commento precedente. Resto in attesa di ulteriori aggiornamenti in merito. io uso Hyundai EU

cdnninja:
I removed requirements file. I don't think we can consume it with an integration? Only pypi

PR #1358, opened 2025-10-18
fix: separate config flow for EU tokens.

cdnninja:
This starts the base required to allow for integration of a better experience for grabbing tokens.   Two sets of code is run once you say where you are.   Translations also move the prompts to the right spot for english.


lbijzeit:
Still no solution for hyundai

PR #1364, opened 2025-10-25
feat: Add buttons and switches to manage actions

payano:
Instead of having actions to handle for example pre-heating, add a switch to turn on heating instead.


custom_components/kia_uvo/button.py
```diff
@@ -0,0 +1,128 @@
...
+            payload,
+            blocking=True,
+            target={"device_id": [self._device.id]},
+        )
+        # refresh current coordinator
+        await self.coordinator.async_request_refresh()
+
+
+# ---- setup ----
+BUTTON_SPECS: list[_ButtonSpec] = [
```
cdnninja: I would think this should follow other standards using a button description.

custom_components/kia_uvo/switch.py
```diff
@@ -0,0 +1,131 @@
...
+
+# ---- helpers ----
+def _vin_from_device(device: DeviceEntry) -> str | None:
+    for dom, ident in device.identifiers:
+        if dom == DOMAIN:
+            return ident
+    return None
+
+
+async def _devices_for_entry(
```
cdnninja: Wouldn't the vehicle get passed into the entity?

custom_components/kia_uvo/switch.py
```diff
@@ -0,0 +1,131 @@
...
+            device,
+            "Climate",
+            "start_climate",
+            "stop_climate",
+            {"duration": 10, "temperature": 21, "climate": True, "defrost": True},
+        )
+        self._attr_icon = "mdi:car-seat-heater"
+
+
+class KiaChargingSwitch(_KiaServiceSwitch):
```
cdnninja: Ideally a description is used not a new class per switch.

custom_components/kia_uvo/switch.py
```diff
@@ -0,0 +1,131 @@
...
+    _attr_assumed_state = True
+
+    def __init__(
+        self,
+        coordinator,
+        device: DeviceEntry,
+        name: str,
+        start_service: str,
+        stop_service: str,
+        start_data: dict[str, Any] | None = None,
```
cdnninja: Where does this data come from?

cdnninja (changes requested):
Thank you for this! Few comments on how this is structured. Some of the comments apply to both button and switch.

cdnninja:
Strings and translations(en.json) also will require updates.

fuatakgun:
thanks a lot for the idea and implementation, and I am sure it works. 
can you be sure that, the implementation follows the existing architecture as we do in sensor? 

given that you are already getting help from AI, you can instruct it to do so

custom_components/kia_uvo/button.py
```diff
@@ -0,0 +1,128 @@
...
+
+async def async_setup_entry(
+    hass: HomeAssistant, entry: ConfigEntry, async_add_entities
+):
+    # ✅ use unique_id per upstream __init__.py
+    coordinator = hass.data[DOMAIN][entry.unique_id]
+    entities: list[KiaServiceButton] = []
+    for device in await _devices_for_entry(hass, entry):
+        for spec in BUTTON_SPECS:
+            entities.append(KiaServiceButton(coordinator, device, spec))
```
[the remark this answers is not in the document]
payano: good question, i have updated the code and the EV specific stuff is guarded by an _is_ev method, hope that works.

custom_components/kia_uvo/switch.py
payano: Code is updated, please take another look.

payano:
Appreciate the great comments on the PR, @cdnninja. The first commit was rough—sorry about that—but I’ve cleaned things up now.

@fuatakgun, mind giving it another look?

It’s not my usual programming language, so thanks for bearing with me.

custom_components/kia_uvo/button.py
```diff
@@ -0,0 +1,215 @@
...
+        icon="mdi:ev-station",
+        service="start_charge",
+        capability=_is_ev,
+    ),
+    KiaButtonDescription(
+        key="stop_charge",
+        name="Stop charge",
+        icon="mdi:ev-station-off",
+        service="stop_charge",
+        capability=_is_ev,
```
cdnninja: maybe a spot to use a lambda for the vehicle.engine_type?

custom_components/kia_uvo/button.py
```diff
@@ -0,0 +1,215 @@
...
+    KiaButtonDescription(
+        key="set_charging_current_60",
+        name="Charging current 60%",
+        icon="mdi:current-ac",
+        service="set_charging_current",
+        payload={"level": 60},
+        capability=_is_ev,
+    ),
+    KiaButtonDescription(
+        key="start_hazard_lights",
```
cdnninja: I think this plus the ones below are only newer cars. Not sure if we have a way to check for that though.   Thoughts?

custom_components/kia_uvo/button.py
```diff
@@ -0,0 +1,215 @@
...
+        async_add_entities(entities)
+
+
+class KiaServiceButton(CoordinatorEntity, ButtonEntity):
+    """Hyundai / Kia Connect button entity class."""
+
+    _attr_has_entity_name = True
+
+    def __init__(
+        self, coordinator, *, vehicle_id: str, description: KiaButtonDescription
```
cdnninja: Typing. Example setup: 

https://github.com/Hyundai-Kia-Connect/kia_uvo/blob/613549543260af38c3551e76e0f5666beb2b3ac1/custom_components/kia_uvo/binary_sensor.py#L325-L340

custom_components/kia_uvo/button.py
```diff
@@ -0,0 +1,215 @@
...
+                        coordinator,
+                        vehicle_id=vehicle_id,
+                        description=desc,
+                    )
+                )
+    if entities:
+        async_add_entities(entities)
+
+
+class KiaServiceButton(CoordinatorEntity, ButtonEntity):
```
cdnninja: HyundaiKiaConnectButton and should inherit HyundaiKiaConnectEntity

cdnninja (changes requested):
Awesome! Looking much better.  Thanks for this.    As a note a strings.json file will be needed as well.

custom_components/kia_uvo/button.py
payano: using this now: `capability=lambda v: _is_electric(v),`

payano:
Thanks for the good comments @cdnninja 👍

payano:
i would have used the enums but i couldn't figure it out how to implement this, i dont like converting enums to strings but this is the best i could come up with:

```
def _is_electric(v: Vehicle) -> bool:
    return str(getattr(et := getattr(v, "engine_type", None), "name", et) or "").lower() in {"ev", "bev", "phev"}

```

cdnninja:
Looks like the engine type ENUM hasn't moved into HA yet,  could either import it from the library or add to constants.    I also realized often we do lean into the data existing so could do that if prefered.   If staying engine type the lambda can be something like:

`lambda vehicle: vehicle.engine_type not in engine_type.ICE`

payano:
Lambda looks better now.

payano:
When you approve, i want to squash my commits before merging 😄

payano:
@cdnninja, do you have any more comments regarding this PR?

custom_components/kia_uvo/button.py
```diff
@@ -0,0 +1,177 @@
...
+    HyundaiKiaButtonEntityDescription(
+        key="set_charging_current_100",
+        name="Charging current 100%",
+        icon="mdi:current-ac",
+        service="set_charging_current",
+        payload={"level": 100},
+        capability=lambda v: v.engine_type != ENGINE_TYPES.ICE,
+    ),
+    HyundaiKiaButtonEntityDescription(
+        key="set_charging_current_60",
```
cdnninja: Should these be number entities?  Same with any that are a value.

cdnninja (commented):
Is a button press the right method for some of these?   Some feel like switches, others have existing types for them.  @fuatakgun  Thoughts?

cdnninja (commented):
Is a button press the right method for some of these?   Some feel like switches, others have existing types for them.  @fuatakgun  Thoughts?

payano:
Yes, please tell me which ones you want to have as switches instead and I'll update the PR.

cdnninja:
Anything we have state for should be a switch if no config required for that item.

vam062313:
@payano any updates on this PR.  Having switches and buttons for this is a great idea.

isarrider:
yeah, I would love that!

payano:
Hello, sorry for such a long delay. Its hard to find time when you have a small baby at home :)

isarrider:
I can relate to the feeling with small kids... ;)
I am looking fwd for this when @cdnninja merges this...

cdnninja:
> I can relate to the feeling with small kids... ;) I am looking fwd for this when @cdnninja merges this...

It still has outstanding feedback so not ready for merge.    Breaking this into smaller PRs would help have it merged sooner as well.

cdnninja:
Similar items have been merged, since no response closing this PR.  Please reopen if you plan to revisit.

Issue #1394, opened 2025-11-25
Lamp status


Dinges28:
**Describe the solution you'd like**
Is it possible to inclusie the lamp status as Home Assistent entiteit?
It is already in the data entity

**Include the logs of the end point from Data in HA if this is a sensor**

**Describe alternatives you've considered**
Extract the attributes as entities with HAXS entity extractor

**Additional context**

JB12134:
I tried looking at these on my EV6 and I couldn't get them to turn on at all, even on the Kia website and app. I don't know if they're a warning that they're burnt out or if it's supposed to tell you they're on. Do you know?

JB12134:
I wrote the code to bring those entities in, as well as a few others that I saw were in my data from Kia that weren't in HA yet (ignition, accessory, sleep mode, transmission condition, remote ignition), but I just want to make sure we name them all right.

Dinges28:
> I tried looking at these on my EV6 and I couldn't get them to turn on at all, even on the Kia website and app. I don't know if they're a warning that they're burnt out or if it's supposed to tell you they're on. Do you know?
> 
> 

As you can see in my example, my left front blinker light is broken, and that shows now "true". It does not change instantly, takes half a day before it is shown

JB12134:
Ah thanks for pointing that out. I can update the names I have to indicate
that.

JB12134:
@Dinges28, are you in Canada? I don't know which countries support this, but my log in Canada looks the same.

Dinges28:
> [@Dinges28](https://github.com/Dinges28), are you in Canada? I don't know which countries support this, but my log in Canada looks the same.

The Netherlands (Europe)

PR #1397, opened 2025-11-28
feat: Add light sensors and other entities from logs to CA and EU

JB12134:
https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/958

https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1394

### DIFF
--- a/custom_components/kia_uvo/binary_sensor.py
+++ b/custom_components/kia_uvo/binary_sensor.py
@@ -303,7 +303,7 @@ SENSOR_DESCRIPTIONS: Final[tuple[HyundaiKiaBinarySensorEntityDescription, ...]] 
     HyundaiKiaBinarySensorEntityDescription(
         key="accessory_on",
         name="Accessory",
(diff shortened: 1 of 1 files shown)


custom_components/kia_uvo/binary_sensor.py
```diff
@@ -300,6 +300,162 @@ class HyundaiKiaBinarySensorEntityDescription(BinarySensorEntityDescription):
         device_class=BinarySensorDeviceClass.WINDOW,
         entity_category=EntityCategory.DIAGNOSTIC,
     ),
+    HyundaiKiaBinarySensorEntityDescription(
+        key="accessory_on",
+        name="Accessory",
+        is_on=lambda vehicle: getattr(vehicle, "accessory_on", None),
```
cdnninja: getattr shouldn't be needed.  The key is checked for a value and the entity is only added if it exists.  Same applies for the others.

cdnninja (changes requested):
Thanks for doing this!

PR #1417, opened 2025-12-11
fix: revert OTP

cdnninja:


cdnninja:
@renatkh reverting the OTP items for now.

PR #1435, opened 2025-12-19
fix: inversion of locked/unlocked state in Homeassistant

Triple-S:
Hello,

I just noticed that Homeassistant uses inverse logic compared to Bluelink for the locked state. In Bluelink `true` means `locked` and `false` means `unlocked` while Homeassistant binary sensors use `true` for `unlocked` and `false` for `locked`. Therefore the state reported by Bluelink needs to be inversed.

Best regards,
Triple-S

### DIFF
--- a/custom_components/kia_uvo/binary_sensor.py
+++ b/custom_components/kia_uvo/binary_sensor.py
@@ -415,7 +415,7 @@ SENSOR_DESCRIPTIONS: Final[tuple[HyundaiKiaBinarySensorEntityDescription, ...]] 
     HyundaiKiaBinarySensorEntityDescription(
         key="is_locked",
         name="Locked",
-        is_on=lambda vehicle: vehicle.is_locked,
+        is_on=lambda vehicle: not vehicle.is_locked,
         device_class=BinarySensorDeviceClass.LOCK,
         entity_category=EntityCategory.DIAGNOSTIC,
     ),


cdnninja:
This needs to be fixed in the api.   Could you open a pr within the library?

Triple-S:
@cdnninja I have to strongly disagree. I think this problem needs to be fixed on the integration side and I would even say it must not be fixed on the API side for the following reason:
The API calls the variable "is_locked" which means it's value answers the question "Is the vehicle locked?" so yes / true for locked and no / false for unlocked and this is all correct. Homeassistent hover asks a different question. It asks "Is the vehicle unlocked?" so it expects yes / true for unlocked and no / false for locked.
Both ways are fine and neither one is right or wrong, its just a matter of converting one system into the other. In other words translate from the API language to the Homeassistent language and in my opinion this needs to be done by the integration and not the API.

cdnninja:
If that is the case something else is wrong here, maybe translation files.   is_locked would be true if locked in the library.  The key for the sensor is the same "is_locked". Are you saying the entity is showing differently?  Is it showing "unlocked" in the interface?

Triple-S:
Yes, the value of the key is correct, if the vehicle is locked, the value of the key is_locked is true as it should be but Homeassistent displays "unlocked". That is because Homeassistent binary sensors have no is_locked or is_unlocked key, they have is_on, so they need to convert from on/off aka. true/false to locked/unlocked, but they do this as if is_on would be is_unlocked which is the opposite of what the API uses.

cdnninja:
Ahh this is caused by device class. https://developers.home-assistant.io/docs/core/entity/binary-sensor/ I will merge.

erikgeurts:
> Ahh this is caused by device class. https://developers.home-assistant.io/docs/core/entity/binary-sensor/ I will merge.

Looking at those docs, I must admit that the way Home Assistant uses On and Off for a lock is counter-intuitive. Having a lock set to "on" when it is actually open is bound to confuse people (as it did here). Good to see it got sorted out in the end!

Triple-S:
@erikgeurts I think its not necessarily counter intuitive but rather a different point of view / a diffrerent way of thinking. I assume the Homeassistent devs designed binary sensors as sort of alarm ligths that turn on when something is in a potential problematic state. An open window is potentially more problematic than a closed one just as an unlocked car is potentially more problematic than a locked one.

erikgeurts:
> @erikgeurts I think its not necessarily counter intuitive but rather a different point of view / a diffrerent way of thinking. I assume the Homeassistent devs designed binary sensors as sort of alarm ligths that turn on when something is in a potential problematic state. An open window is potentially more problematic than a closed one just as an unlocked car is potentially more problematic than a locked one.

Agreed, but the terms could have been more intuitive. For a device class like "problem" or "running" it's fairly obvious what 'on' refers to. The device class for a lock could have been made more intuitive by calling it "locked", and then a state of 'on' would be immediately obvious. Similarly ambiguous device classes are 'door' or 'window'. I think the confusion stems from the fact that some device classes are named after the 'thing', whereas others are named after the 'state of the thing'. For example 'connectivity' could have been called 'connected' and then 'on' would have been immediately obvious.

PR #1468, opened 2026-01-08
fix: Improve error handling for sleeping vehicle telematics

rp1783:
Add graceful KeyError handling when the vehicle's telematics module (TMU) is asleep and the API returns error 502 without a vehicleStatus key.

This issue commonly occurs with older vehicles (especially 2020 models) when the TMU goes to sleep after ~4 hours of inactivity. The Hyundai BlueLink API returns a 502 error response without the expected vehicleStatus field, causing a KeyError that crashes the integration.

Changes:
- Catch KeyError exceptions during force update and cached update attempts
- Provide helpful warning messages explaining the sleeping vehicle issue
- Suggest opening the BlueLink app to wake the vehicle
- Skip the failed update cycle and return existing data
- Change log level from ERROR to WARNING since this is expected behavior

The integration will automatically retry on the next scheduled update cycle, and will work normally once the vehicle wakes up (via driving, opening the app, or successful force refresh).

Fixes issues where users see:
- Integration fails to load with "Config Not Ready" error
- KeyError: 'vehicleStatus' in logs
- Only 4 entities created instead of full set

Tested with: 2020 Hyundai Sonata (USA region)

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -129,6 +129,15 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
                     self.vehicle_manager.check_and_force_update_vehicles,
                     self.force_refresh_interval,
                 )
+            except KeyError as key_error:
+                _LOGGER.warning(
+                    f"API response missing expected key '{key_error}'. "
+                    f"This usually means the vehicle's telematics module is asleep. "
+                    f"Opening the Hyundai BlueLink app on your phone will wake it up. "
+                    f"Skipping this update cycle and will retry on the next scheduled update."
+                )
+                # Don't retry cached update if we're getting malformed responses
+                return self.data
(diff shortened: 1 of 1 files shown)


cdnninja:
How would this solve only 4 sensors. If values = none on startup the sensor isn't added which means it will never show.   On bootup we need a successful connection or need to go into a retry state.   While making solution more clear via exception is good we do want startup to stop.

josh-m-sharpe:
Hmm I also encountered a lot of 502 errors - usually when I'm trying to develop/work on my car-related home assistant dashboards and I restart home assistant a lot.  This triggers triggers an update to the API whenever that happens.

For that reason I assumed I was getting rated limited, so I built this PR to rate limit the update request: https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/1470

What makes you think it's this TMU-gone-to-sleep thing?

cdnninja:

PR #1470, opened 2026-01-08
fix: Add circuit breaker pattern for API rate limit handling

josh-m-sharpe:
Fixes issue where Genesis API goes into rate limit mode, then the plugin attempts to call the update method every minute indefinitely.

- Track consecutive failures with timestamps
- Enter degraded mode after 3 failures in 5 minutes
- 15-minute cooldown to let API recover
- Never raise UpdateFailed to prevent retry storms
- Always return stale data instead of crashing
- Automatic recovery when API responds
- Enhanced logging with vehicle identification


cdnninja:
I am not comfortable with monkey patches on our own solution.    This also reaches into the API layer instead of just vehicle manager.  

I think separate PRs will also help review.

PR #1476, opened 2026-01-11
feat: save and recall token

cdnninja:
https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/972

Save and recall tokens to allow for no repeating prompts for OTP.


cdnninja:
Will add via the USA pull request #1479

PR #1505, opened 2026-01-19
fix: convert heating to correct type

com6056:
Noticed I couldn't get the "Rear Defrost & Accessories" climate function enabled in requests on my 2025 EV6 in the USA, dug in a bit and found `heating` wasn't being converted properly to an `int` which meant these weren't matching: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/84bd2276e24ec97bd672d864c423d99808cb53fb/hyundai_kia_connect_api/KiaUvoApiUSA.py#L878-L879

Tested with this change and it works for me now 🎉

### DIFF
--- a/custom_components/kia_uvo/services.py
+++ b/custom_components/kia_uvo/services.py
@@ -87,6 +87,8 @@ def async_setup_services(hass: HomeAssistant) -> bool:
         steering_wheel = call.data.get("steering_wheel")
 
         # Confirm values are correct datatype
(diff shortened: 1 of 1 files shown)


com6056:
@cdnninja any chance you can give this a review when you have a moment? Thanks! 🙏

cdnninja:
Looks good to me! Thank you.

PR #1551, opened 2026-02-25
feat: Retain values after refresh failures

mvroosmalen1970:
Added error handling for DeviceIDError during token refresh to maintain data availability.


cdnninja:
I believe this directly breaks. Home assistant standards.   We shouldn't show data with a current time stamp when it's not true.

Issue #1538, opened 2026-02-16
Error in API which results in unavailability of all entities


jee67:
Please check Services, Known Bug / Issues and Troubleshooting over here first: https://github.com/fuatakgun/kia_uvo/blob/master/README.md
**Region and Brand of car**
Europe - KIA
**Version of the integration**
2.50.0
**Describe the bug**
Very frequently the error reported in the log below occurs. All entities become unavailable. 

**Debug logs if an error occurred**
```
Deze fout is ontstaan door een aangepaste integratie.

Logger: custom_components.kia_uvo.coordinator
Bron: helpers/update_coordinator.py:416
integratie: Hyundai / Kia Connect (documentatie, problemen)
Eerst voorgekomen: 14:54:15 (1 gebeurtenis)
Laatst gelogd: 14:54:15

Unexpected error fetching kia_uvo data
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 416, in _async_refresh
    self.data = await self._async_update_data()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/kia_uvo/coordinator.py", line 115, in _async_update_data
    await self.async_check_and_refresh_token()
  File "/config/custom_components/kia_uvo/coordinator.py", line 177, in async_check_and_refresh_token
    await self.hass.async_add_executor_job(
        self.vehicle_manager.check_and_refresh_token
    )
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.13/site-packages/hyundai_kia_connect_api/VehicleManager.py", line 213, in check_and_refresh_token
    result = self.api.refresh_access_token(
        self.token,
    )
  File "/usr/local/lib/python3.13/site-packages/hyundai_kia_connect_api/ApiImpl.py", line 335, in refresh_access_token
    return self.login(
           ~~~~~~~~~~^
        username=token.username, password=token.password, pin=token.pin
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 188, in login
    device_id = self._get_device_id(stamp)
[...]

PR #1552, opened 2026-02-26
fix(coordinator): convert transient API errors to UpdateFailed

tamcore:
Kia's EU API (prd.eu-ccapi.kia.com) intermittently returns a DeviceIDError ('Invalid request body - service problem, because of anything problem') or a ReadTimeoutError during token refresh (ref: https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1538) These are transient server-side failures with no action required from the user.

Before this change, these exceptions propagated through async_check_and_refresh_token() as unexpected exceptions into the DataUpdateCoordinator. HA's coordinator counts consecutive unexpected exceptions and cancels the config entry entirely after enough failures, making all entities permanently unavailable until the integration is manually reloaded.

While this will not fix the upstream issues with Kia's API, this should improve error handling and make it less painful for the user and should close https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1538.

This PR proposes 2 improvements:

1. _async_update_data: add an except-all clause after AuthenticationError, that converts any remaining exception from token refresh into UpdateFailed. UpdateFailed is the documented way to signal a recoverable API error: entities become temporarily unavailable and the coordinator retries automatically after 60 seconds (retry_after=60) instead of waiting for the next full poll interval. async_check_and_refresh_token itself is simplified to have no try/except — all exception handling is centralised in the caller.

2. All action methods (lock, unlock, start/stop charge, climate, etc.): wrap the executor call in try/except and raise HomeAssistantError with a descriptive message. Previously any failure here propagated silently with no feedback to the user or the HA frontend.

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -16,7 +16,7 @@ from hyundai_kia_connect_api import (
 )
 from hyundai_kia_connect_api.exceptions import AuthenticationError
 
-from homeassistant.exceptions import ConfigEntryAuthFailed
+from homeassistant.exceptions import ConfigEntryAuthFailed, HomeAssistantError
 
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import (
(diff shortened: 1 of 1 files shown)


cdnninja (commented):
Thanks for this! Overall looks good. One comment.

PR #1563, opened 2026-03-03
feat: add reconfigure flow (re-auth or set PIN)

tamcore:
Adds a reconfigure option to the integration (`Settings -> Integrations -> Kia Uvo / Hyundai Bluelink -> 3dot menu -> Reconfigure`) with two choices:

- `Re-authenticate your account`: full login flow and updates the existing credentials
- `Set / change PIN`: password-type input to set/update the PIN required for remote commands (lock/unlock/climate), without requiring full re-authentication

**Disclaimer**: Inspiration has been taken from https://github.com/andreadegiovine/homeassistant-stellantis-vehicles, since the Reconfigure option there saved me a few times already, and manually hacking the pin into `.storage/core.config_entries` just because it might've been forgotten on first log-in is potentially error prone.

**Disclaimer No.2**: AI has been used to provide the translations.

### DIFF
--- a/custom_components/kia_uvo/config_flow.py
+++ b/custom_components/kia_uvo/config_flow.py
@@ -15,6 +15,7 @@ import voluptuous as vol
 
 from homeassistant import config_entries
 from homeassistant.config_entries import ConfigEntry
+from homeassistant.helpers.selector import selector
 from homeassistant.const import (
     CONF_PASSWORD,
     CONF_PIN,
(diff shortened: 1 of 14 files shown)


cdnninja:
Thanks for this!! 

Looks like pre commit is failing.  Could you confirm this has been tested locally?

tamcore:
Fixed. Sorry for the oversight. The fresh clone of the repo didn't have the pre-commit hook installed. Besides that, both reconfigure flows have been tested on my Home Assistant instance.

Issue #1059, opened 2025-03-11
Connection Error, ConnectionResetbyPeer


memattyg:
Please check Services, Known Bug / Issues and Troubleshooting over here first: https://github.com/fuatakgun/kia_uvo/blob/master/README.md
**Region and Brand of car**
USA, KIA
**Version of the integration**
2.36.1
**Describe the bug**
I will inconsistently, but consistently get the following error when issuing a command to the car:

Error requesting kia_uvo data: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))

**Debug logs if an error occurred**

Outlined here: https://github.com/fuatakgun/kia_uvo/blob/master/README.md

**To Reproduce**
Inconsistently occurs, have not found a pattern.
**Expected behavior**
No error, execute command
**Screenshots**

**Additional context**

bstempien:
I am seeing the same issue with my US EV6 trying to do a climate start.

lemisieur:
Same here, in Canada trying to do a climate start

cdnninja:
Log would be helpful.

bstempien:
March 14, 2025 at 7:56:09 AM
Error: ('Connection aborted.', ConnectionResetError(104, 'Connection reset
by peer'))
Result:
params:
  domain: kia_uvo
  service: start_climate
  service_data:
    duration: 10
    climate: true
    heating: '0'
    device_id: ee4e570008031224198e3118ac074d8d
    temperature: 69
  target: {}
running_script: false

cdnninja:
> Same here, in Canada trying to do a climate start

To confirm you have the same connection reset error?

lemisieur:
> > Same here, in Canada trying to do a climate start
> 
> To confirm you have the same connection reset error?

Correct, let me try to get some logs for you

andrew-manger:
Getting the connection reset error too, just passively

```
This error originated from a custom integration.

Logger: custom_components.kia_uvo.coordinator
Source: helpers/update_coordinator.py:393
integration: Hyundai / Kia Connect (documentation, issues)
First occurred: 01:11:14 (2 occurrences)
Last logged: 02:11:15

Error requesting kia_uvo data: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
```

emilio2003:
im getting the same error in Canada

emilio2003:
Any solutions?

cdnninja:
[...]

PR #1568, opened 2026-03-10
fix: validate both charge limits before sending API call

### LINKED ISSUE #1569: Charge limit sliders send invalid values and snap back after setting
## Description

The charge limit sliders (AC and DC) fail to work correctly in two ways:

1. **Sending invalid values:** The Kia API requires both AC and DC charge limits in a single compound call. When a user moves one slider, the integration reads the "other" cached value from the vehicle object. If that value is `None` (e.g. before a force refresh has populated it), the API silently ignores the request — the slider appears to work but nothing changes on the car.

2. **Snapping back to stale values:** After successfully setting charge limits, the integration does a soft refresh (cmm/gvi) which does not return `targetSOC` for some vehicles (e.g. 2020 Kia Niro EV). The slider snaps back to the old value because the soft refresh doesn't pick up the new charge limits.

## Steps to Reproduce

1. Install the integration on a 2020 Kia Niro EV (USA)
2. After a soft refresh (no force refresh yet), observe that charge limit sliders show values but `ev_charge_limits_ac/dc` may be `None` internally
3. Try to move one slider — the API call silently fails because the other value is `None`
4. After a force refresh populates both values, move a slider — it sets correctly but snaps back to the old value within seconds because the post-set soft refresh doesn't include `targetSOC`

## Expected Behavior

- If the other charge limit is not available, show a clear error message telling the user to force refresh first
- After setting charge limits, the slider should reflect the new value (via force refresh read-back)

## Vehicle

- 2020 Kia Niro EV (USA, KiaUvoApiUSA)
- cmm/gvi response does NOT include `targetSOC` for this vehicle

## Fix

PR #1568 (depends on API fix in Hyundai-Kia-Connect/hyundai_kia_connect_api#1059)

kingpanther13:
Closes #1569

## Summary

Fixes charge limit sliders sending invalid values and snapping back to stale data after being set.

**Root cause:** The Kia API requires both AC and DC charge limits in a single compound call. When a user moves one slider, the integration reads the "other" cached value from the vehicle object. If that value is `None` (e.g. before a force refresh has been performed), the API silently ignores the request. Additionally, after setting charge limits, the soft refresh (cmm/gvi) does not return `targetSOC` for some vehicles, causing sliders to snap back to stale values.

### Changes

**number.py — Validate before sending:**
- Added `_is_valid_charge_limit()` to check the cached "other" charge limit is a valid integer 50-100 in steps of 10
- If invalid, raises `HomeAssistantError` with a user-friendly message telling them to force refresh first or set the other slider manually
- V2L code is completely untouched

**coordinator.py — Force refresh after setting:**
- Added `async_await_action_and_force_refresh()` method that does a force refresh (rems/rvs) instead of soft refresh (cmm/gvi) after the action completes
- Force refresh is wrapped in `try/except` so a failed refresh still notifies HA entities to update (prevents entities from being stuck in stale state)
- Uses `async_set_updated_data()` instead of `async_refresh()` to avoid a redundant cmm/gvi API call — the force refresh already updates vehicle objects in-place, so we just notify HA to re-read state
- `async_set_charge_limits()` now uses the new method so fresh charge limit values are read back immediately
- Prevents slider snap-back on vehicles where cmm/gvi doesn't return targetSOC (e.g. 2020 Kia Niro EV)

### Companion PR

> **API side:** https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1059
> Fixes force refresh to always update targetSOC with fresh data, adds robust parsing and 18 unit tests.
>
> **⚠️ Merge order: PR #1059 (API) must be merged FIRST**, then this PR (#1568).

## Test plan

- [x] Manual test: moving AC slider when DC is unknown shows error toast in HA UI
- [x] Manual test: after force refresh, both sliders show correct values and can be set
- [x] Manual test: after setting a charge limit, slider reflects new value within ~60 seconds (force refresh reads it back)
- [ ] Verify V2L slider (if present) still works normally
- [x] Verify no regression for other vehicle commands (lock, unlock, climate, etc.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -213,6 +213,11 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
         Used after setting charge limits because the soft refresh (cmm/gvi)
         does not return targetSOC for some vehicles. A force refresh (rems/rvs)
         ensures the fresh charge limits are read back immediately.
+
+        Uses async_set_updated_data instead of async_refresh to avoid a
+        redundant cmm/gvi API call — the force refresh already updates the
+        vehicle objects in-place (rems/rvs + cmm/gvi), so we just need to
+        notify HA entities to re-read their state.
         """
         try:
             await asyncio.sleep(5)
@@ -224,10 +229,15 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
                 60,
             )
         finally:
-            await self.hass.async_add_executor_job(
-                self.vehicle_manager.force_refresh_vehicle_state, vehicle_id
-            )
-            await self.async_refresh()
+            try:
+                await self.hass.async_add_executor_job(
+                    self.vehicle_manager.force_refresh_vehicle_state, vehicle_id
+                )
+            except Exception:
+                _LOGGER.exception(
+                    "Force refresh after setting charge limits failed"
+                )
+            self.async_set_updated_data(self.data)
 
     async def async_lock_vehicle(self, vehicle_id: str):
         await self.async_check_and_refresh_token()
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -207,6 +207,28 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
         finally:
             await self.async_refresh()
 
+    async def async_await_action_and_force_refresh(self, vehicle_id, action_id):
+        """Wait for action then force refresh to get fresh vehicle data.
+
+        Used after setting charge limits because the soft refresh (cmm/gvi)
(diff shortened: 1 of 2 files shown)


kingpanther13:
@cdnninja could you look into merging this one please? I had it on my dev/test version of the addon but the latest update just pushed me back to the live version, and now my charge limits aren't working again, they're just showing unknown.

kingpanther13:
@cdnninja sorry for another ping. I just realized that the current version of this repo/integration is still pinned to API 4.7.2,  this PR won't work till the API is at least 4.7.3 which is the one that incorporates https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1059 . Looks like current API version is 4.8.2

cdnninja:
Are you able to create a pr to bump the api version?

kingpanther13:
Done. #1593

PR #1570, opened 2026-03-10
feat: Add button and switch entities for vehicle controls on device page

### LINKED ISSUE #1566: EV charging entity not present
Please check Services, Known Bug / Issues and Troubleshooting over here first: https://github.com/fuatakgun/kia_uvo/blob/master/README.md
**Region and Brand of car**
USA 2020 Kia Niro EV 
**Version of the integration**
2.52
**Describe the bug**
 I get a total of 32 entires but the ones for starting and stopping ev charging are not present.  They are defined the the YAML file but do not show up on Home Assistant 

**Debug logs if an error occurred**

Outlined here: https://github.com/fuatakgun/kia_uvo/blob/master/README.md

**To Reproduce**

**Expected behavior**
able to control ev charging 

**Screenshots**

**Additional context**

kingpanther13:
Closes #1566

## Summary

Exposes vehicle actions that were previously only accessible via service calls as proper HA entities on the device page. Users currently have to create custom dashboard buttons calling services directly to control EV charging and climate — these should be native entities.

### New entities

| Entity | Type | Availability | Description |
|--------|------|-------------|-------------|
| Force Refresh | Button | All vehicles | Triggers a force refresh (rems/rvs) to get fresh vehicle data |
| EV Charging | Switch | EV vehicles only | Start/stop EV charging with on/off state tracking (`ev_battery_is_charging`) |
| Climate | Switch | All vehicles | Start/stop climate control with on/off state tracking (`air_control_is_on`). Start uses API defaults (70°F, 5 min, AC on, defrost off) |

### Files changed

- **`button.py`** (new) — `ForceRefreshButton` entity using `ButtonEntity`
- **`switch.py`** (new) — `EVChargingSwitch` and `ClimateSwitch` entities using `SwitchEntity`
- **`__init__.py`** — Registers `Platform.BUTTON` and `Platform.SWITCH` in the PLATFORMS list

### Design decisions

- **Switch vs Button for charging/climate:** Switches are the correct HA entity type because these actions have persistent on/off state tracked via existing vehicle attributes
- **Climate defaults:** `ClimateRequestOptions()` with no args is intentional — `KiaUvoApiUSA.start_climate()` fills sensible defaults (70°F, 5 min, climate=True, defrost=False). Users who want specific settings can still use the `start_climate` service call
- **EV Charging guard:** Only added for vehicles with `ev_battery_percentage` (same guard as charge limit sliders in `number.py`)
- **Force Refresh scope:** Refreshes all vehicles under the coordinator (same as existing `force_update` service)
- **No API changes needed:** All coordinator methods already exist; these are thin entity wrappers

## Test plan

- [ ] Verify Force Refresh button appears on device page and triggers a force refresh
- [ ] Verify EV Charging switch appears for EV vehicles and shows correct charging state
- [ ] Verify EV Charging switch can start and stop charging
- [ ] Verify Climate switch appears for all vehicles and shows correct climate state
- [ ] Verify Climate switch can start and stop climate control
- [ ] Verify no new entities appear for non-EV vehicles that shouldn't have them
- [ ] Verify existing entities (sensors, binary sensors, lock, charge limit sliders) still work

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/custom_components/kia_uvo/button.py
+++ b/custom_components/kia_uvo/button.py
@@ -1,0 +1,51 @@
+"""Button for Hyundai / Kia Connect integration."""
+
+from __future__ import annotations
+
+import logging
+
+from homeassistant.components.button import ButtonEntity
+from homeassistant.config_entries import ConfigEntry
+from homeassistant.core import HomeAssistant
+from homeassistant.helpers.entity_platform import AddEntitiesCallback
+
+from hyundai_kia_connect_api import Vehicle
+
+from .const import DOMAIN
+from .coordinator import HyundaiKiaConnectDataUpdateCoordinator
+from .entity import HyundaiKiaConnectEntity
+
+_LOGGER = logging.getLogger(__name__)
+
+
+async def async_setup_entry(
+    hass: HomeAssistant,
+    config_entry: ConfigEntry,
+    async_add_entities: AddEntitiesCallback,
+) -> None:
+    coordinator = hass.data[DOMAIN][config_entry.unique_id]
+    entities = []
+    for vehicle_id in coordinator.vehicle_manager.vehicles.keys():
+        vehicle: Vehicle = coordinator.vehicle_manager.vehicles[vehicle_id]
+        entities.append(ForceRefreshButton(coordinator, vehicle))
+
+    async_add_entities(entities)
+    return True
+
+
+PARALLEL_UPDATES = 1
+
+
+class ForceRefreshButton(ButtonEntity, HyundaiKiaConnectEntity):
+    def __init__(
+        self,
+        coordinator: HyundaiKiaConnectDataUpdateCoordinator,
+        vehicle: Vehicle,
+    ):
+        HyundaiKiaConnectEntity.__init__(self, coordinator, vehicle)
+        self._attr_unique_id = f"{DOMAIN}_{vehicle.id}_force_refresh"
+        self._attr_name = f"{vehicle.name} Force Refresh"
+        self._attr_icon = "mdi:refresh"
+
+    async def async_press(self) -> None:
+        await self.coordinator.async_force_update_all()
--- a/custom_components/kia_uvo/button.py
+++ b/custom_components/kia_uvo/button.py
@@ -39,9 +39,7 @@ async def async_setup_entry(
     for vehicle_id in coordinator.vehicle_manager.vehicles.keys():
         vehicle: Vehicle = coordinator.vehicle_manager.vehicles[vehicle_id]
         for description in BUTTON_DESCRIPTIONS:
-            entities.append(
-                HyundaiKiaConnectButton(coordinator, description, vehicle)
-            )
+            entities.append(HyundaiKiaConnectButton(coordinator, description, vehicle))
 
     async_add_entities(entities)
 
--- a/custom_components/kia_uvo/button.py
+++ b/custom_components/kia_uvo/button.py
@@ -30,7 +30,6 @@ async def async_setup_entry(
         entities.append(ForceRefreshButton(coordinator, vehicle))
 
     async_add_entities(entities)
-    return True
 
 
 PARALLEL_UPDATES = 1
--- a/custom_components/kia_uvo/button.py
+++ b/custom_components/kia_uvo/button.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from dataclasses import dataclass
 import logging
 from typing import Final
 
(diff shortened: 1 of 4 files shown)


custom_components/kia_uvo/button.py
```diff
@@ -0,0 +1,50 @@
...
+        coordinator: HyundaiKiaConnectDataUpdateCoordinator,
+        vehicle: Vehicle,
+    ):
+        HyundaiKiaConnectEntity.__init__(self, coordinator, vehicle)
+        self._attr_unique_id = f"{DOMAIN}_{vehicle.id}_force_refresh"
+        self._attr_name = f"{vehicle.name} Force Refresh"
+        self._attr_icon = "mdi:refresh"
+
+    async def async_press(self) -> None:
+        await self.coordinator.async_force_update_all()
```
cdnninja: This would make the button on the car level update all cars.   Maybe we take this change to expose `force_refresh_vehicle_state` from the API?

cdnninja (changes requested):
Thank you for this!  Feedback attached.

custom_components/kia_uvo/button.py
kingpanther13: Good catch! Fixed in a0a1ce6.

- Added `async_force_refresh_vehicle(vehicle_id)` to the coordinator, which calls the existing `vehicle_manager.force_refresh_vehicle_state(vehicle_id)` for per-vehicle refresh
- The Force Refresh button now only refreshes its own vehicle
- Uses `async_set_updated_data` instead of `async_refresh` to avoid a redundant cached API call after the force refresh

custom_components/kia_uvo/switch.py
```diff
@@ -0,0 +1,94 @@
...
+
+from hyundai_kia_connect_api import ClimateRequestOptions, Vehicle
+
+from .const import DOMAIN
+from .coordinator import HyundaiKiaConnectDataUpdateCoordinator
+from .entity import HyundaiKiaConnectEntity
+
+_LOGGER = logging.getLogger(__name__)
+
+EV_CHARGING_KEY = "ev_battery_is_charging"
```
cdnninja: I think these can move into the definition.

cdnninja (changes requested):
Moving in the right direction!

custom_components/kia_uvo/switch.py
kingpanther13: Fixed in 5228249 — removed the separate key constants. Keys are now defined directly in the `HyundaiKiaSwitchDescription` instances.

kingpanther13:
I can add more stuff in if you'd like for the climate controls, I frankly didn't realize there was more than a basic climate on and off till I started building this. However I don't have the subscription so I don't think I will be able to test the climate stuff reliably. Occasionally the scripts work for me for basic on and off, but usually I get an email that says it failed. The charging scripts have always worked though

cdnninja (changes requested):
Agreed on keeping it small and building the structure. More can be added easily if we get the structure right!

custom_components/kia_uvo/switch.py
```diff
@@ -0,0 +1,95 @@
...
+        self._attr_unique_id = f"{DOMAIN}_{vehicle.id}_{self._key}"
+        self._attr_icon = description.icon
+        self._attr_name = f"{vehicle.name} {description.name}"
+
+    @property
+    def is_on(self) -> bool | None:
+        return getattr(self.vehicle, self._key, None)
+
+    async def async_turn_on(self, **kwargs) -> None:
+        await getattr(self.coordinator, self._description.on_action)(self.vehicle.id)
```
cdnninja: Why is this a getattr command?  Should we do a lambda?   Here is a google example for the other integration I support: 

https://github.com/home-assistant/core/blob/4efb10dae12affa57344a33937b903e1715ae879/homeassistant/components/vesync/switch.py#L203-L210   would be a bit simpler since it doesn't return messages in the same way etc. 

https://github.com/home-assistant/core/blob/4efb10dae12affa57344a33937b903e1715ae879/homeassistant/components/vesync/switch.py#L106-L113

kingpanther13: Good call — switched to typed Callable lambdas on the description (`value_fn`, `exists_fn`, `on_fn`, `off_fn`) following the vesync pattern you linked. No more `getattr` string lookups. Also moved the entity existence checks into `exists_fn` per description, which cleans up `async_setup_entry`. Pushed in a53b1b9.

PR #1578, opened 2026-03-16
fix: EV entity discovery when initial EV fields are null

lortiejp:
## Summary

This fixes a discovery issue for EV/PHEV vehicles where EV-related entities can be skipped permanently if the backend returns `null` EV fields during initial setup.

## Problem

Entity creation currently depends on values already being present:

- `sensor.py` only creates entities when `getattr(vehicle, description.key, None) is not None`
- `binary_sensor.py` does the same for EV binary sensors
- `number.py` only creates AC/DC charging limit entities when `ev_battery_percentage` is present

For at least some Kia Canada EV vehicles, the backend may initially omit or null out `status.evStatus.*`. When that happens, EV entities are never created, even if valid EV data appears on a later refresh.

## Fix

- Create EV/PHEV sensor entities based on `vehicle.engine_type`, even when the current EV value is `None`
- Create EV/PHEV binary sensors the same way
- Create AC/DC charge limit numbers for electrified vehicles even if `ev_battery_percentage` is missing
- Normalize `0/1` backend flags for EV charging / plugged-in binary sensors to proper booleans

## Validation

Tested locally with a Kia EV6 in Canada.

Before the patch, EV charge-related entities were missing. After reloading the integration with this patch, the EV entities were created and the values matched the Kia app.

### DIFF
--- a/custom_components/kia_uvo/binary_sensor.py
+++ b/custom_components/kia_uvo/binary_sensor.py
@@ -40,7 +40,7 @@ def _is_electrified_vehicle(vehicle: Vehicle) -> bool:
 
 
 def _should_add_binary_sensor(
-    description: "HyundaiKiaBinarySensorEntityDescription", vehicle: Vehicle
+    description: HyundaiKiaBinarySensorEntityDescription, vehicle: Vehicle
 ) -> bool:
     """Create EV entities even when the backend currently reports null values."""
     if getattr(vehicle, description.key, None) is not None:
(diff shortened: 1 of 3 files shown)


cdnninja:
When the data is missing is it due to differences between force refresh and cached?

custom_components/kia_uvo/binary_sensor.py
```diff
@@ -26,6 +27,28 @@
 _LOGGER = logging.getLogger(__name__)
 
 
+def _normalize_optional_bool(value: object) -> bool | None:
```
cdnninja: Seems this should be fixed in the library if situations exist where it isn't a bool.

lortiejp:
I'm not sure whether it's due to these differences. At installation of the integration, the data was missing and the entities were missing. Then, refresh didn't create them either. I tried both cached and forced refresh.

cdnninja:
Please review debug logs, that will tell.  This could just be a mapping issue on one of them.

cdnninja:

PR #1585, opened 2026-03-29
feat: expose Brazil controls, diagnostics, and trip data

eduardobittencourt:
## Summary
- expose the working Brazil remote features in Home Assistant with climate, windows, start engine, richer diagnostics, and trip data entities
- improve BR distance and trip presentation so the integration reflects the backend data in kilometers
- keep the local integration aligned with the patched Brazil API behavior that was validated live

## What Changed
- enabled the `climate`, `cover`, and new `button` platforms for the BR flow
- added a windows cover entity and an explicit `Start Engine` button
- made climate setup resilient when BR snapshots do not include an initial temperature or unit
- surfaced additional BR diagnostics such as accessory state, lamp faults, remote control availability, oil warning, hazard lights, and tail lamps
- added automatic BR trip and notification history fetches in the coordinator
- exposed monthly and daily trip entities plus derived trip metrics for dashboard use
- ensured trip and distance sensors prefer kilometers instead of drifting to miles in Home Assistant

## Why
The Brazil backend already supports more status and command coverage than the integration was exposing. The goal here is to make the Home Assistant side reflect the real BR capabilities that were validated with a live account, while keeping the UI practical for day-to-day use.

## Validation
- `python3 -m py_compile custom_components/kia_uvo/__init__.py custom_components/kia_uvo/binary_sensor.py custom_components/kia_uvo/button.py custom_components/kia_uvo/climate.py custom_components/kia_uvo/coordinator.py custom_components/kia_uvo/cover.py custom_components/kia_uvo/sensor.py`
- live validation on a BR Home Assistant instance for login, windows, climate, trip sensors, diagnostics, and the explicit start engine control

## Notes
This PR was prepared with AI assistance, so it is being opened as a draft intentionally. I will still review the changes manually and polish anything needed before marking it ready for review.


custom_components/kia_uvo/coordinator.py
```diff
@@ -105,6 +106,56 @@ def __init__(self, hass: HomeAssistant, config_entry: ConfigEntry) -> None:
             ),
         )
 
+    def _update_extended_br_data(self) -> None:
```
cdnninja, line 109: I am not comfortable with region specific code.    All should be generic.

cdnninja (changes requested):
Please split this into smaller PRs.   One per platform change.   We shouldn't need to reference a region in kia_uvo.

PR #1603, opened 2026-04-12
fix: Entity naming for Home Assistant 2026.4+ 

### LINKED ISSUE #1584: Entity names get triple-prefixed after HA 2026.4 friendly_name migration
## Description

After upgrading to Home Assistant 2026.4, all entity friendly names are triple-prefixed with the vehicle name. For example:

- **Before:** `KONA Air Conditioner`
- **After:** `KONA (KONA) KONA Air Conditioner`

This affects all entities (sensors, binary sensors, locks, device trackers, numbers, etc.).

## Root cause

HA 2026.4 changed how `friendly_name` is composed — the device name is now consistently prepended to entity names. The kia_uvo integration currently:

1. Sets the **device name** to `f"{vehicle.name} ({vehicle.model})"` in `entity.py:24`, which becomes `KONA (KONA)` when name and model are identical
2. Sets **entity names** with `f"{vehicle.name} {description.name}"` (e.g. in `sensor.py`, `binary_sensor.py`, etc.), producing `KONA Air Conditioner`
3. Does **not** set `has_entity_name = True`

Because `has_entity_name` is not set, HA 2026.4 prepends the full device name to the entity name, resulting in `KONA (KONA) KONA Air Conditioner`.

## Suggested fix

1. Set `has_entity_name = True` on `HyundaiKiaConnectEntity`
2. Remove the `{vehicle.name}` prefix from `_attr_name` in all entity types — let HA compose it from the device name automatically
3. When `vehicle.name == vehicle.model`, set device name to just `vehicle.name` instead of `f"{vehicle.name} ({vehicle.model})"` to avoid the `KONA (KONA)` duplication

## Environment

- **HA version:** 2026.4.0b2
- **kia_uvo version:** latest from HACS
- **Vehicle:** Hyundai KONA (EU)

mastameista:
## Summary
Adjust the integration to Home Assistant's 2026.4+ entity naming model so vehicle entities no longer end up with duplicated name prefixes on newer HA versions.

## Changes
- set `_attr_has_entity_name = True` on the shared base entity
- stop prefixing entity names with `vehicle.name`
- keep the vehicle name on the device only
- register the device with separate `name` and `model` values instead of using `"{name} ({model})"` for both

## Why
With recent Home Assistant versions, the device name is used when composing friendly names for entities that have `has_entity_name = True`.

This integration was still embedding the vehicle name directly into each entity name while also exposing a device name like `"{name} ({model})"`, which could lead to names such as:

- `<vehicle name> (<vehicle model>) <vehicle name> Air Conditioner`

After this change, the registry and UI use the expected split:

- device name: `<vehicle.name>`
- device model: `<vehicle.model>`
- entity original name: `Air Conditioner`
- friendly name: `<vehicle.name> Air Conditioner`

This better matches Home Assistant's naming model than storing `"{name} ({model})"` in both device fields.

## Compatibility
This should not change existing `unique_id`s and should not require entity ID migration for existing installs.

For new installs, entity IDs should remain in the same shape as before, for example:
- `binary_sensor.<vehicle_name_slug>_air_conditioner`
- `sensor.<vehicle_name_slug>_last_updated_at`
- `lock.<vehicle_name_slug>_door_lock`

## Verification
- validated locally with `py -m compileall custom_components\\kia_uvo`
- deployed to a Home Assistant dev instance
- confirmed updated device/entity registry values and corrected friendly names there

Fixes #1584.

### DIFF
--- a/custom_components/kia_uvo/binary_sensor.py
+++ b/custom_components/kia_uvo/binary_sensor.py
@@ -490,7 +490,7 @@ class HyundaiKiaConnectBinarySensor(BinarySensorEntity, HyundaiKiaConnectEntity)
         super().__init__(coordinator, vehicle)
         self.entity_description: HyundaiKiaBinarySensorEntityDescription = description
         self._attr_unique_id = f"{DOMAIN}_{vehicle.id}_{description.key}"
-        self._attr_name = f"{vehicle.name} {description.name}"
+        self._attr_name = description.name
         if description.entity_category:
             self._attr_entity_category = description.entity_category
 
--- a/custom_components/kia_uvo/climate.py
+++ b/custom_components/kia_uvo/climate.py
@@ -90,7 +90,7 @@ class HyundaiKiaCarClimateControlSwitch(HyundaiKiaConnectEntity, ClimateEntity):
         )
         self.vehicle_manager = coordinator.vehicle_manager
         self._attr_unique_id = f"{DOMAIN}_{vehicle.id}_climate_control"
-        self._attr_name = f"{vehicle.name} Climate Control"
+        self._attr_name = self.entity_description.name
 
         # set the Climate Request to the current actual state of the car
         self.climate_config = ClimateRequestOptions(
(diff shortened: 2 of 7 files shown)


cdnninja:
Thank you for this! Fingers crossed it doesn't drive issue.

Issue #1571, opened 2026-03-11
Cannot select Region when setting up


rogerholifield7:
Cannot select Region when setting up. Nothing gets selected when I click on region entry.

Please check Services, Known Bug / Issues and Troubleshooting over here first: https://github.com/fuatakgun/kia_uvo/blob/master/README.md
**Region and Brand of car**

**Version of the integration**

**Describe the bug**

**Debug logs if an error occurred**

Outlined here: https://github.com/fuatakgun/kia_uvo/blob/master/README.md

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

Libellenreiter:
Same problem for me. 
I think it has something to do with changes in homeassistant 2026.2 oder 2026.3 to the dropdown menus. 
Some of the custom components I use shipped updates mentioned such changes.

nodepoints:
Dammit - I've been having the same issue trying to get kia_uvo running again after a re-install.  Cannot select Region drop down.

yorb:
Same issue here. The Region dropdown appears, but is not filled when choosing an option. This causes the next screen to ask for a token, which I don't have. :(

<img width="609" height="483" alt="Image" src="https://github.com/user-attachments/assets/6b19e192-49a4-4e47-9e8a-e0843c34315e" />

cdnninja:
> Same issue here. The Region dropdown appears, but is not filled when choosing an option. This causes the next screen to ask for a token, which I don't have. :(
> 
> <img width="609" height="483" alt="Image" src="https://github.com/user-attachments/assets/6b19e192-49a4-4e47-9e8a-e0843c34315e" />

What region are you in?  If you are EU that means you got through this screen successfully.   Maybe just doesn't show an item is selected?

yorb:
> What region are you in? If you are EU that means you got through this screen successfully. Maybe just doesn't show an item is selected?

USA. And you're right, I swear this wasn't working before but now if I choose USA from the Region selector, and then click Submit, it takes me to the correct form. I did just update to 2026.3.2, maybe that or just the restart allowed it to work?

At any rate, it still doesn't show the value, but at least it works now!

PR #1622, opened 2026-04-17
refactor: translation mechanism to eliminate hardcoded English (#1)

blka:
I recently added translations for Polish, but I didn’t realise at all that files in the ‘translations’ folder aren’t being taken into account. This PR fixes that
  
  Summary

  Enable Home Assistant's modern entity translation system (translation_key) for all 80+ entities across 8 platform files. Entity names were previously hardcoded in English within Python code, making translations invisible in the UI regardless of language files existing.

  Problem

  The integration had translation files (pl.json, de.json, etc.) covering config flow and services, but all entity names were hardcoded in English via name= in EntityDescription classes. The _attr_has_entity_name = True flag was already set in the base entity class, signaling intent to use the modern naming system, but translation_key was never used. As a
  result:

  - Entity names always displayed in English regardless of HA language setting
  - Entities with device_class (BATTERY, DISTANCE, TIMESTAMP) fell back to HA's generic built-in translations ("Bateria", "Odległość", "Znacznik czasu") instead of showing descriptive custom names
  - Entities without a device_class showed only the device name (e.g. just "Santa Fe") since no name was available
  - Manual self._attr_name assignments (e.g. f"{vehicle.name} {description.name}") caused duplicate device name prefixes in entity IDs

  Changes

  Python files — entity translation mechanism

  - sensor.py: Added translation_key to all 35 SensorEntityDescription entries, removed all hardcoded name=. Changed self._description to self.entity_description (required for HA to locate translation_key). Replaced name property on VehicleEntity, DailyDrivingStatsEntity, TodaysDailyDrivingStatsEntity with _attr_translation_key.
  - binary_sensor.py: Added translation_key to all 47 HyundaiKiaBinarySensorEntityDescription entries, removed all name=. Removed self._attr_name = description.name.
  - button.py: Added translation_key="force_refresh", removed name="Force Refresh". Removed manual self._attr_name = f"{vehicle.name} {description.name}" (HA auto-prepends device name with _attr_has_entity_name = True). Changed self._description to self.entity_description.
  - switch.py: Added translation_key to 2 switch descriptions, removed name= and _attr_name prepend.
  - climate.py: Added translation_key="climate_control", removed name="Climate Control" and _attr_name assignment.
  - number.py: Added translation_key to 3 number descriptions, removed name= and _attr_name. Changed self._description to self.entity_description throughout.
  - lock.py: Replaced self._attr_name = "Door Lock" with self._attr_translation_key = "door_lock".
[...]

Issue #1591, opened 2026-04-06
EU region cannot be selected during Hyundai token generation


SLS-Saarlodri:
### What happened?

During the Hyundai token generation process, it is not possible to select a region.
The script/tool does not offer the EU region, so login/token generation cannot be completed correctly.

### Expected behavior

There should be an option to select the EU region during the setup/token generation process for Hyundai accounts.

### Actual behavior

No EU region can be selected.
Because of that, the Hyundai login/token generation does not work as expected.

### Steps to reproduce

1. Start the Hyundai token generation script/tool
2. Go through the login/setup process
3. Try to select the region
4. Notice that EU cannot be selected

### Environment

- Brand: Hyundai
- Platform: iPad / browser
- Account region: EU
- Vehicle/App region: EU
- Problem: no region selection available

### Additional context

I already tried again with the latest version, but the EU region still cannot be selected.
Because of this, the integration/token setup cannot be completed.

If needed, I can provide screenshots of the login/setup flow where the region selection is missing.

PR #1623, opened 2026-04-17
fix: initialization of dropdown Region menu

blka:
should be merged after translations PR from https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/1622

⏺ Summary

  Fix the config flow dropdown for Region and Brand selection so the chosen value persists in the UI after clicking.

image

image

⏺ Problem

  When adding the integration, the Region and Brand dropdowns use vol.In(REGIONS) / vol.In(BRANDS) with integer-keyed dicts ({1: "Europe", 2: "Canada", ...}). The HA frontend cannot properly handle integer values in dropdown fields — after selecting an option, the dropdown closes and the field reverts to "Region*" / "Brand*" instead of showing the selected
  value. This forces users to guess whether their selection was accepted.

⏺ Changes

  custom_components/kia_uvo/config_flow.py:
  - Replaced vol.In(REGIONS) and vol.In(BRANDS) with selector({"select": {"options": [{"value": str(k), "label": v}], "mode": "dropdown"}}) — the modern HA selector API that correctly renders dropdowns with explicit value/label pairs
  - Added int() conversion for CONF_REGION and CONF_BRAND after form submission — selectors return string values ("1", "2", ...) while VehicleManager and dict lookups require integers
  - Removed unused STEP_USER_DATA_SCHEMA — dead code combining username/password/region/brand in a single step that was never referenced

⏺  Test plan

  - Start adding the integration in HA UI
  - Select a region from the dropdown — verify the selected value stays visible in the field
  - Select a brand from the dropdown — verify the selected value stays visible in the field
  - Complete the config flow — verify the integration receives correct integer values for region/brand
  - Test reconfigure flow — verify PIN change and re-auth still work


blka:
fixes:
https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1571
https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1591

cdnninja:
Thank you for this! Are you able to rebase to make it easier to see what is changing on this?   It should just be region items.

PR #1624, opened 2026-04-17
Attempt Silent Reauthentication/Re-Login Before Triggering Manual Reauth

### LINKED ISSUE #1581: try old credentials to get a token before failing and requiring user intervention
**Is your feature request related to a problem? Please describe.**
The integration opens a repair every few days, maybe once every week or two? asking to re enter credentials. Doing so fixes things again for a while. Credentials haven’t changed. 

I’m using the Hyundai API in the USA. 

**Describe the solution you'd like**
Before opening a new repair, try to get a new token with the credentials I used last time. It’s all the same data (region, brand, username, password, pin) I used in the initial setup. 

**Describe alternatives you've considered**
Existing solution works through the repair workflow but is tedious and requires unnecessary user intervention when credentials were already given to the integration and haven’t changed.

TheOneOgre:
## Summary

This changes the auth refresh flow so the integration does not immediately trigger Home Assistant reauth/repair when an auth session expires but the stored credentials are still valid.

Instead, the integration now:

- when token refresh fails for a US account, attempts one immediate silent login with stored credentials; invalid credentials still trigger reauth immediately, while temporary backend failures retry later with backoff (30 minutes, then 6 hours)
- only falls back to `ConfigEntryAuthFailed` if the silent login also fails

Resolves #1581 

## Why

The config entry already stores the auth inputs needed for recovery, including:

- region
- brand
- username
- password
- PIN
- token

Before this change, an expired session could still escalate directly to Home Assistant’s manual reauth flow even when a normal login with stored credentials would have succeeded.

## Behavior Change

Before:
- token/session refresh failure could immediately raise `ConfigEntryAuthFailed`
- Home Assistant would prompt manual repair/reauth
- temporary service outages could be treated like auth failures

After:
- Account/token refresh failures first attempt one silent re-login
- invalid credentials still go to reauth immediately
- maintenance/service-down responses back off and retry later
- retry backoff is staged at 30 minutes, then 6 hours

## Scope

- keeps the change localized to the coordinator auth path
- does not change the user-facing config flow

## Testing

Used focused coordinator tests covering:
- Hyundai USA silent relogin after refresh auth failure
- non-US entries do not attempt silent relogin (I cannot test external regions, but there isn't a reason this shouldn't be applicable to that auth flow as well)
- invalid credentials still fail immediately
- maintenance/service-unavailable auth failures use retry backoff
- successful auth resets the retry backoff

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -14,6 +14,7 @@ from hyundai_kia_connect_api import (
     ScheduleChargingClimateRequestOptions,
     Token,
 )
+from hyundai_kia_connect_api.ApiImpl import OTPRequest
 from hyundai_kia_connect_api.exceptions import AuthenticationError
 
 from homeassistant.exceptions import ConfigEntryAuthFailed, HomeAssistantError
@@ -31,6 +32,9 @@ from homeassistant.helpers.update_coordinator import DataUpdateCoordinator, Upda
 from homeassistant.util import dt as dt_util
 
 from .const import (
+    BRANDS,
+    BRAND_GENESIS,
+    BRAND_HYUNDAI,
     CONF_BRAND,
     CONF_FORCE_REFRESH_INTERVAL,
     CONF_NO_FORCE_REFRESH_HOUR_FINISH,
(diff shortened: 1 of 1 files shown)


cdnninja:
I provided a comment in the other thread. I think part of this logic is duplicated in the library.  This region seems to potentially not be throwing the correct exceptions for the system to handle it. 

Check and refresh token in the library is called before all actions so should be used to relogin.

cdnninja:
I merged some code on the library side that may help this issue.  Here is the retry login that already exists in the integration:

https://github.com/Hyundai-Kia-Connect/kia_uvo/blob/0b723b967bdd32dd7d6de27160355bd18b2755bc/custom_components/kia_uvo/coordinator.py#L118-L132

60 second retry for recoverable failures.    We could extend that further if needed.  Or increment it each time it is used.  First step those it get USA Hyundai to actually use it.

Issue #1620, opened 2026-04-16
Force Update = Duplicate request


peter4200:
### Region

EU

### Brand of Car

Hyundai

### Version of the Integration

2.54.0

### Describe the Bug

In my integration I have two cars. One is deactivated (because I sold it) and the other is my current car (Ioniq 5 N-Line MY2025).
When I select a forced update on the active car in the developer tools, I get the error "Duplicate request"

### Debug Logs (if an error occurred)

2026-04-16 18:03:55.208 DEBUG (SyncWorker_6) [hyundai_kia_connect_api.KiaUvoApiEU] hyundai_kia_connect_api - Received forced vehicle data: {'retCode': 'F', 'resCode': '4004', 'resMsg': 'Duplicate request - Duplicate request', 'msgId': 'de966640-39ad-11f1-bda4-63c21cbcdfd5'}
2026-04-16 18:03:55.210 ERROR (MainThread) [homeassistant.helpers.script.websocket_api_script] websocket_api script: Error executing script. Unexpected error for call_service at pos 1: Duplicate request - Duplicate request
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 531, in _async_step
    await getattr(self, handler)()
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 1018, in _async_step_call_service
    response_data = await self._async_run_long_action(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<9 lines>...
    )
    ^
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 631, in _async_run_long_action
    return await long_task
           ^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2817, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2860, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/kia_uvo/services.py", line 68, in async_handle_force_update
    await coordinator.async_force_update_all()
  File "/config/custom_components/kia_uvo/coordinator.py", line 185, in async_force_update_all
    await self.hass.async_add_executor_job(
        self.vehicle_manager.force_refresh_all_vehicles_states
    )
[...]

Issue #1617, opened 2026-04-15
Force Update button from integration makes all entities unknown


HappyPlotter:
### Region

EU

### Brand of Car

Hyundai

### Version of the Integration

2.54.0

### Describe the Bug

When pressing the "Force update" button from the integration, most of the entities become "unknown" and never get a new updated value.

### Debug Logs (if an error occurred)

2026-04-15 20:55:50.126 DEBUG (SyncWorker_4) [hyundai_kia_connect_api.KiaUvoApiEU] hyundai_kia_connect_api - Received forced vehicle data: {'retCode': 'S', 'resCode': '0000', 'resMsg': {'ServiceNo': 'RVS-B', 'svcTime': 0, 'FncCnt': 0, 'RetCode': 'S', 'resCode': '200', 'SID': '568d2a10-38f4-11f1-871f-d94b2f21b826', 'svcSID': '568d2a10-38f4-11f1-871f-d94b2f21b826'}, 'msgId': '568d2a10-38f4-11f1-871f-d94b2f21b826'}
2026-04-15 20:55:50.533 DEBUG (SyncWorker_4) [hyundai_kia_connect_api.KiaUvoApiEU] hyundai_kia_connect_api - _get_location response: {'retCode': 'S', 'resCode': '0000', 'resMsg': {'coord': {'lat': xx.654844, 'lon': yy.31748, 'alt': 0, 'type': 0}, 'head': 207.6999969482422, 'speed': {'value': 0, 'unit': 0}, 'time': '20260415205128'}, 'msgId': '56c669f0-38f4-11f1-871f-330da02a0ed2'}
2026-04-15 20:55:50.533 DEBUG (SyncWorker_4) [hyundai_kia_connect_api.KiaUvoApiEU] hyundai_kia_connect_api - SOC Levels couldn't be found. May not be an EV.

### To Reproduce

Press the "update" button inside the integration

### Expected Behavior

All the entities should be populated

### Screenshots

<img width="1170" height="751" alt="Image" src="https://github.com/user-attachments/assets/a034b6c6-e464-4e7b-8747-0d87b9de5171" />

### Additional Context

When the action for force update is executed from developer tools it returns all the entities but seems to be running the forced and then the cached function. The forced function returns nothing as above and the cached returns the data.

[...]

PR #1629, opened 2026-04-21
fix: reduce duplicate request errors and improve post-action refresh handling for set charge limits

jablo:
Tested on EU Hyundai Ioniq 5 (2025) and Kia EV3 (2025).

Calling force_refresh_vehicle_state() immediately after previous requests can trigger "Duplicate request" (4004) errors from the backend.

Adding a small delay before force refresh reduces the frequency of these failures and makes manual force updates more reliable. This appears to be related to backend timing/rate limiting rather than incorrect request parameters.

After set_charge_limits, the current flow uses force_refresh_vehicle_state(). In this situation, the force refresh sequence can fail in _get_location() with "Duplicate request" (4004), and may also return incomplete data.

In my testing, the updated charge limits become available via normal cached refresh after a short delay (~40 seconds), though possibly this may vary depending on backend timing. This change replaces the post-action force refresh with a delayed normal refresh.

This avoids duplicate request errors and provides stable retrieval of updated charge limits without additional API load.

Related to:
- Hyundai-Kia-Connect/kia_uvo#1620
- Hyundai-Kia-Connect/kia_uvo#1617
- https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1102

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -190,6 +190,7 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
     async def async_force_refresh_vehicle(self, vehicle_id: str) -> None:
         """Force refresh a single vehicle's state."""
         await self.async_check_and_refresh_token()
+        await asyncio.sleep(20)
         await self.hass.async_add_executor_job(
             self.vehicle_manager.force_refresh_vehicle_state, vehicle_id
         )
@@ -238,6 +239,7 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
             )
         finally:
             try:
+                await asyncio.sleep(60)
                 await self.hass.async_add_executor_job(
                     self.vehicle_manager.force_refresh_vehicle_state, vehicle_id
                 )
@@ -350,6 +352,7 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
     async def async_set_charge_limits(self, vehicle_id: str, ac: int, dc: int):
         await self.async_check_and_refresh_token()
         try:
+            await asyncio.sleep(5)
             action_id = await self.hass.async_add_executor_job(
                 self.vehicle_manager.set_charge_limits, vehicle_id, ac, dc
             )
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -214,6 +214,7 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
                 60,
             )
         finally:
+            await asyncio.sleep(40)
             await self.async_refresh()
 
     async def async_await_action_and_force_refresh(self, vehicle_id, action_id):
@@ -359,7 +360,7 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
         except Exception as err:
             raise HomeAssistantError(f"Failed to set charge limits: {err}") from err
         self.hass.async_create_task(
-            self.async_await_action_and_force_refresh(vehicle_id, action_id)
+            self.async_await_action_and_refresh(vehicle_id, action_id)
         )
 
     async def async_set_charging_current(self, vehicle_id: str, level: int):


custom_components/kia_uvo/coordinator.py
```diff
@@ -213,6 +214,7 @@ async def async_await_action_and_refresh(self, vehicle_id, action_id):
                 60,
             )
         finally:
+            await asyncio.sleep(40)
```
cdnninja, line 217: Seems this should be added to the library for apis that don't properly support the check action cycle.  Otherwise this delays the proper check for regions that may support that.   Not sure if right now the check action works but rather not break that function.

custom_components/kia_uvo/coordinator.py
```diff
@@ -350,13 +353,14 @@ async def async_stop_charge(self, vehicle_id: str):
     async def async_set_charge_limits(self, vehicle_id: str, ac: int, dc: int):
         await self.async_check_and_refresh_token()
         try:
+            await asyncio.sleep(5)
```
cdnninja, line 356: Shouldn't this be after setting not before?   Check and refresh token doesn't interact with the car.  Only a login if expired otherwise nothing.

custom_components/kia_uvo/coordinator.py
```diff
@@ -190,6 +190,7 @@ async def async_force_update_all(self) -> None:
     async def async_force_refresh_vehicle(self, vehicle_id: str) -> None:
         """Force refresh a single vehicle's state."""
         await self.async_check_and_refresh_token()
+        await asyncio.sleep(20)
```
cdnninja, line 193: Typically check token doesn't interact with the api.  Does this one solve a specific issue?

cdnninja (commented):
Thanks for this!

custom_components/kia_uvo/coordinator.py
jablo: Two comments: 1) The 40 seconds was my first working guess, up from initial 20 seconds. However 40 seconds seem to not always be enough, and current I'm using 60 seconds - that seems much more stable.

2) This is my first dive into any HA integration, so I'm still finding my way around. I see what you mean, and definitely it should go in the API. In my I current experiment, I have done so by implementing check_action_status() as simply a "time.sleep(60)" in the EU api. Do you think that is reasonable?

custom_components/kia_uvo/coordinator.py
jablo: It is hard to debug how these delays interact. I'll experiment some more.

custom_components/kia_uvo/coordinator.py
jablo: You are right. I made this change before the get_location commit some days ago. Triggering force_update from my HA UI now gives duplicate request on the get_location, with or without. I'll try and remove this, add a delay in force_refresh in the API and update the PR. 

2026-04-22 10:21:03.171 DEBUG (SyncWorker_1) [hyundai_kia_connect_api.KiaUvoApiEU] hyundai_kia_connect_api - Received forced vehicle data: {'retCode': 'S', 'resCode': '0000', 'resMsg': {'ServiceNo': 'RVS-B', 'svcTime': 0, 'FncCnt': 0, 'RetCode': 'S', 'resCode': '200', 'SID': '33a2d450-3e24-11f1-bda4-23b0de3316dc', 'svcSID': '33a2d450-3e24-11f1-bda4-23b0de3316dc'}, 'msgId': '33a2d450-3e24-11f1-bda4-23b0de3316dc'}
2026-04-22 10:21:03.340 DEBUG (SyncWorker_1) [hyundai_kia_connect_api.KiaUvoApiEU] hyundai_kia_connect_api - _get_location response: {'retCode': 'F', 'resCode': '4004', 'resMsg': 'Duplicate request - Duplicate request', 'msgId': '33bb1740-3e24-11f1-bda4-823645ed292b'}
2026-04-22 10:21:03.340 ERROR (SyncWorker_1) [hyundai_kia_connect_api.KiaUvoApiEU] hyundai_kia_connect_api - _get_location failed: Duplicate request - Duplicate request
Traceback (most recent call last):
  File "/config/dev/hyundai_kia_connect_api/KiaUvoApiEU.py", line 861, in _get_location
    _check_response_for_errors(response)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/config/dev/hyundai_kia_connect_api/ApiImplType1.py", line 93, in _check_response_for_errors
    raise error_code_mapping[response["resCode"]](response["resMsg"])
hyundai_kia_connect_api.exceptions.DuplicateRequestError: Duplicate request - Duplicate request

cdnninja:
Based on a few other items that may need tinkering maybe revise this pr to just cover switching to a cached call for set charge limits?   I can merge that quick which should help for some people.

jablo:
I will try and do that as quickly as I can --- it just turns out my daily life has turned a bit tumultuous at the moment. I hope to get time to sit at my computer one of the coming days =:/

> Based on a few other items that may need tinkering maybe revise this pr to just cover switching to a cached call for set charge limits? I can merge that quick which should help for some people.

cdnninja:
Hopefully okay, I took the liberty of updating the one call.   I gave credit to you in the commit.

blka:
Thanks for investigating the DuplicateRequestError issue! This is a real problem that affects users sending consecutive commands.

However, I think the hardcoded sleep approach has some downsides:
- The delay values (20/40/60s) are arbitrary — on slow connections or busy servers, they may still not be enough
- They add latency to every action, even when there's no conflict
- As cdnninja noted, throttling logic belongs in the API library layer

PR #1635 takes a different approach: an `asyncio.Lock` that serializes all vehicle actions at the coordinator level. When a second action is sent while the first is still in progress, it's rejected immediately with a clear message ("Another vehicle action is in progress") instead of failing with a cryptic API error. This avoids unnecessary delays while still preventing overlapping API calls.

The one change in this PR that *is* valuable on its own is switching `set_charge_limits` from force refresh to cached refresh — and I see cdnninja already incorporated that.

jablo:
> Hopefully okay, I took the liberty of updating the one call. I gave credit to you in the commit.

You are most welcome, and thank you. The Kia and Hyundai apps seemed to become stable last week and I have re-enabled my HA integration again in the week-end --- just to be certain that HA was not at fault for the instabilities I saw.

jablo:
> As cdnninja noted, throttling logic belongs in the API library layer

I certainly agree -- the idea with posting my changes was to show my analyses/findings. Arbitrary delays is a stop-gap at best.

I'll rescind my PRs and hope to contribute something else at another time.

jablo:
Superseded by other PRs

PR #1635, opened 2026-04-28
feat: add _async_send_action helper with action lock and UnsupportedControlError handling

blka:
## Summary

   - Add `_async_send_action` helper method that centralizes the vehicle action lifecycle: token refresh → send action → poll for completion → refresh data
   - Add `asyncio.Lock` (`_action_lock`) to serialize vehicle actions, preventing `DuplicateRequestError` (resCode 4004) when commands overlap
   - Reject concurrent actions immediately with a clear user-facing message ("Another vehicle action is in progress") instead of blocking or failing with a cryptic API error
   - Catch `UnsupportedControlError` (resCode 4005) and convert to a user-friendly `HomeAssistantError` ("Vehicle does not support this action") instead of surfacing the raw API error
   - Change action polling from fire-and-forget (`async_create_task`) to `await` so the lock is held for the entire action lifecycle (send → poll → refresh)
   - Refactor all 16 action methods to delegate to `_async_send_action` via lambdas, eliminating duplicated try/except blocks

   ## Problems fixed

   1. **DuplicateRequestError (4004)** — Sending two vehicle commands in quick succession (e.g., open windows then close windows) caused the Hyundai API to reject the second with resCode 4004. Root cause: `async_create_task` allowed overlapping API calls. Fix: action lock serializes commands.

   2. **UnsupportedControlError (4005)** — HEV vehicles report read-only status data (e.g., `air_control_is_on`) but don't support certain control actions (e.g., `stop_climate`). The 4005 error code was unmapped, surfacing as a raw `APIError`. Fix: map 4005 to `UnsupportedControlError` and convert to a clear `HomeAssistantError`.

   3. **Code duplication** — All 16 action methods repeated the same try/except pattern. The `_async_send_action` helper consolidates this into one place.

   ## Requires

   - `hyundai_kia_connect_api` >= version with `UnsupportedControlError` (see [PR: Hyundai-Kia-Connect/hyundai_kia_connect_api](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1115))

   ## Test plan

   - [ ] Send two vehicle commands in quick succession (e.g., lock then unlock) → second should get "Another vehicle action is in progress" message
   - [ ] Call `stop_climate` on an HEV vehicle → should get "Vehicle does not support this action" instead of raw 4005 error
   - [ ] Verify all 16 action methods still work correctly (lock, unlock, climate, charge, windows, etc.)
   - [ ] Verify action polling completes fully before the next action can start (lock is held for entire lifecycle)
   - [ ] Check HA logs for no unexpected errors


custom_components/kia_uvo/manifest.json
```diff
@@ -8,6 +8,6 @@
   "iot_class": "cloud_polling",
   "issue_tracker": "https://github.com/Hyundai-Kia-Connect/kia_uvo/issues",
   "loggers": ["kia_uvo", "hyundai_kia_connect_api"],
-  "requirements": ["hyundai_kia_connect_api==4.10.4"],
+  "requirements": ["hyundai_kia_connect_api==4.11.0"],
```
cdnninja: Are you able to remove this bump? It has been done and may occur again which this would role back.

custom_components/kia_uvo/coordinator.py
```diff
@@ -245,52 +249,71 @@ async def async_await_action_and_force_refresh(self, vehicle_id, action_id):
...
 
-    async def async_lock_vehicle(self, vehicle_id: str):
-        await self.async_check_and_refresh_token()
-        try:
-            action_id = await self.hass.async_add_executor_job(
-                self.vehicle_manager.lock, vehicle_id
+    async def _async_send_action(
+        self,
+        vehicle_id: str,
+        action_fn,
```
cdnninja: Are you able to add type?

custom_components/kia_uvo/coordinator.py
```diff
@@ -245,52 +249,71 @@ async def async_await_action_and_force_refresh(self, vehicle_id, action_id):
...
+        error_label: str,
+        *,
```
cdnninja: What is the * for?

custom_components/kia_uvo/coordinator.py
```diff
@@ -245,52 +249,71 @@ async def async_await_action_and_force_refresh(self, vehicle_id, action_id):
...
+        force_refresh: bool = False,
+    ):
+        """Send a vehicle action, wait for completion, and refresh data.
+
+        Serializes actions with a lock to prevent DuplicateRequestError
+        from the Hyundai API when commands overlap. If another action is
+        already in progress, raises HomeAssistantError immediately so
+        the user gets a clear message instead of a mysterious long wait.
+        """
+        if self._action_lock.locked():
```
cdnninja, line 270: Does this actually solve anything?   In my experience we aren't sending duplicate calls but rather we don't know if the last call is done since the API just returns certain requests a sync.  Root cause is "check_action_status" in some libraries isn't implemented.

custom_components/kia_uvo/coordinator.py
```diff
@@ -245,52 +249,71 @@ async def async_await_action_and_force_refresh(self, vehicle_id, action_id):
...
+            await self.async_check_and_refresh_token()
+            try:
+                action_id = await self.hass.async_add_executor_job(action_fn)
+            except UnsupportedControlError as err:
+                raise HomeAssistantError(
+                    f"Vehicle does not support this action: {err}"
+                ) from err
+            except Exception as err:
+                raise HomeAssistantError(f"Failed to {error_label}: {err}") from err
+            if force_refresh:
```
cdnninja: Should this maybe move up into the try catch? It would let us throw exceptions and correctly handle them for the refresh.

custom_components/kia_uvo/coordinator.py
cdnninja, line 270: To add to this.   Lets ensure our native OOTB function is working.  https://developers.home-assistant.io/docs/integration_fetching_data#request-parallelism 

I think it is.

cdnninja (changes requested):
Thanks for doing this!

blka:
Thanks for the thorough review! I've rebased on master and addressed all your points:

**1. Version bump in manifest.json** — Removed. The rebase picked up the upstream v4.11.0 bump, so our changes no longer touch manifest.json.

**2. Type hint on `action_fn`** — Added: `action_fn: Callable[[], Any]` with proper imports (`Callable` from `collections.abc`, `Any` from `typing`).

**3. The `*` keyword-only separator** — Removed. With only one optional parameter after it, the `*` wasn't adding much clarity. `force_refresh` is now a regular keyword argument.

**4. `asyncio.Lock` vs HA's request parallelism** — I looked into this carefully. HA's `PARALLEL_UPDATES = 1` creates a **per-platform** semaphore, so it serializes actions within a single platform (e.g., two lock presses). However, a lock press and a climate command from **different** platforms would still execute concurrently because they acquire separate semaphores. The Hyundai API rejects **any** overlapping commands with `DuplicateRequestError`, regardless of which entity type triggered them, so we need a coordinator-level lock. That said, you're right that this doesn't cover the case where `check_action_status` isn't properly implemented in the API library — that's a separate root cause. The lock addresses cross-platform concurrency; the `check_action_status` gap is an API library issue.

**5. Polling/refresh in try/catch** — Done. The action-send phase and the polling/refresh phase now have separate try/except blocks. Polling failures are logged but not propagated to the user, since the action itself was already sent successfully. The regular update cycle will eventually pick up the new state.

**Other changes from rebase:**
- `async_set_charge_limits` now uses regular refresh (not force refresh), respecting your change from `a5e5fec`
- `async_await_action_and_force_refresh` still exists (upstream code) and is called when `force_refresh=True`, but charge limits no longer uses it per your direction

blka:
For context: PR #1629 is attempting to fix the same DuplicateRequestError issue using hardcoded `asyncio.sleep()` delays (20/40/60s) before API calls. This PR (#1635) is the proper fix — the action lock serializes commands without unnecessary delays and gives users a clear error message instead of a mysterious API failure.

The only useful change from #1629 is switching `set_charge_limits` from force refresh to cached refresh, which cdnninja has already incorporated separately.

blka:
Just tested the `_async_send_action` pattern from our PR #1665 (which implements the same action lock approach) on a real Hyundai Santa Fe (EU). The fix works great — rapid button presses now get a clear, user-facing error in the Home Assistant UI instead of a cryptic API error.

Here's the log showing the behavior after the fix:

```
2026-05-19 11:03:05.397 DEBUG [hyundai_kia_connect_api.ApiImplType1] Set Navigation Request: {'deviceID': '...', 'poiInfoList': [{'name': 'Dom', 'coord': {'lat': 50.0647, 'lon': 19.945, ...}}]}
2026-05-19 11:03:05.845 DEBUG [hyundai_kia_connect_api.ApiImplType1] Set Navigation Response: {'retCode': 'S', 'resCode': '0000', 'msgId': '8c6d9140-...'}  ← first press succeeds

2026-05-19 11:03:07.993 WARNING [custom_components.kia_uvo.coordinator] Vehicle action 'set navigation' rejected: another action is already in progress
2026-05-19 11:03:07.994 ERROR [homeassistant.components.websocket_api.http.connection] Error during service call to kia_uvo.set_navigation: Another vehicle action is in progress. Please wait for it to complete and try again.

2026-05-19 11:03:08.740 WARNING [custom_components.kia_uvo.coordinator] Vehicle action 'set navigation' rejected: another action is already in progress
2026-05-19 11:03:08.742 ERROR [homeassistant.components.websocket_api.http.connection] Error during service call to kia_uvo.set_navigation: Another vehicle action is in progress. Please wait for it to complete and try again.

2026-05-19 11:03:09.224 WARNING [custom_components.kia_uvo.coordinator] Vehicle action 'set navigation' rejected: another action is already in progress
2026-05-19 11:03:09.225 ERROR [homeassistant.components.websocket_api.http.connection] Error during service call to kia_uvo.set_navigation: Another vehicle action is in progress. Please wait for it to complete and try again.
```

Before the fix, these overlapping calls would hit the API and get `DuplicateRequestError` (resCode 4004). After the fix:
1. **First press** — sends the command, polls for completion, refreshes data. The lock is held for the entire lifecycle.
2. **Rapid subsequent presses** — immediately rejected at the coordinator level with a clear message, never reaching the API. The user sees a proper toast notification in the HA UI.

PR #1643, opened 2026-05-07
feat: add AC charging current limit as number entity

blka:
## Summary

Adds `ev_charging_current` as a **number entity** (slider, 60–100%, step 10) for EV AC charging current limit configuration.

Replaces the initial `select` platform approach (hardcoded 60/90/100% dropdown) with a number entity that matches the pattern used by the existing AC/DC charging limit and V2L limit number entities.

### Changes

- **`number.py`**: Add `EV_CHARGING_CURRENT_KEY` (`ev_charging_current`) to `NUMBER_DESCRIPTIONS`
  - Min 60%, max 100%, step 10%, unit `%`
  - `async_set_native_value` calls `coordinator.async_set_charging_current`
- **`select.py`**: REMOVED — no longer needed
- **`__init__.py`**: Remove `Platform.SELECT` from platform list
- **`strings.json`** + **`translations/en.json`**: Move translation from `select` to `number` section

### Why number over select

Per @cdnninja's review feedback: a number entity with a slider is more flexible than a hardcoded dropdown. The API accepts integer values, so a number entity (with min/max/step constraints) is a more natural fit than a select with three fixed options.

Independent PR — no dependencies on other PRs.

## Test plan

- [ ] Install on EV vehicle with charging current support and verify number entity appears with slider (60–100%, step 10)
- [ ] Test sliding to different charging current values
- [ ] Verify the number entity sends the correct integer value to the API
- [ ] Install on HEV vehicle and verify no charging current entity is created


cdnninja:
What is the value on adding a select? I believe a number entity with steps for this already exists.

blka:
You're right — there's already a `number` entity pattern in this integration. Looking at the existing sensor, `ev_charging_current` is a sensor (percentage) and there's a `set_charging_current` service. A `number` entity with `min: 60`, `max: 100`, `step: 10` would be more natural than a `select` with 3 hardcoded options.

I'll rework this as a `number` entity instead. Should I update this PR or close it and open a new one?

cdnninja:
That feels like an AI answer. What's your intent on this pr?

blka:
Ah, true. I didn’t catch that. Closing that out.

PR #1644, opened 2026-05-07
feat: enable climate platform with None guard

blka:
## Summary

Enables the `climate` platform (previously commented out) with proper None guards and UnsupportedControlError handling.

### Changes

- **`climate.py`**: 
  - Add `exists_fn` — `vehicle.air_control_is_on is not None` — prevents entity creation on vehicles without climate state
  - Add `UnitOfTemperature.CELSIUS` fallback when `_air_temperature_unit` is `None`
  - Catch `UnsupportedControlError` in `async_set_hvac_mode` and `async_set_temperature` — some vehicles (e.g. HEV) report climate state but don't support remote climate control commands. The error is logged as a warning instead of causing an unhandled exception
  - Use `vehicle.air_control_is_on.values()` for cleaner iteration
  - Fix docstring ("Set up climate platform" instead of "binary_sensor platform")
- **`__init__.py`**: Uncomment `Platform.CLIMATE`

Depends on: #1641 (None guard on `air_control_is_on`)

### Known limitation

Some vehicles report climate state (`air_control_is_on`) but don't support remote climate control commands (e.g. HEV vehicles via the EU API). On these vehicles, the climate entity will appear and display the current state, but attempting to change the HVAC mode or temperature will log a warning instead of crashing. This is a limitation of the vehicle/API, not the integration.

### Testing

- `air_control_is_on=0` — climate entity correctly created
- Attempting to start climate on HEV triggers `UnsupportedControlError` — now caught and logged as warning
- Without the None guard, vehicles without climate support would cause errors

## Test plan

- [ ] Install on vehicle with climate support and verify climate entity appears
- [ ] Test start/stop climate operations on supported vehicle
- [ ] Install on HEV vehicle and verify climate entity appears (read-only state)
- [ ] Verify attempting climate control on unsupported vehicle shows warning in logs (not crash)
- [ ] Verify temperature unit fallback works when `_air_temperature_unit` is None

### DIFF
--- a/custom_components/kia_uvo/climate.py
+++ b/custom_components/kia_uvo/climate.py
@@ -6,6 +6,7 @@ import logging
 from time import sleep
 
 from hyundai_kia_connect_api import ClimateRequestOptions, Vehicle, VehicleManager
+from hyundai_kia_connect_api.exceptions import UnsupportedControlError
 
 from homeassistant.components.climate import ClimateEntity, ClimateEntityDescription
 from homeassistant.components.climate.const import (
@@ -202,19 +203,22 @@ class HyundaiKiaCarClimateControlSwitch(HyundaiKiaConnectEntity, ClimateEntity):
     async def async_set_hvac_mode(self, hvac_mode):
         """Set the operation mode of the in-car climate control."""
 
-        if hvac_mode == HVACMode.OFF:
-            await self.hass.async_add_executor_job(
-                self.vehicle_manager.stop_climate,
-                self.vehicle.id,
-            )
-            self.vehicle.air_control_is_on = False
-        else:
-            await self.hass.async_add_executor_job(
-                self.vehicle_manager.start_climate,
-                self.vehicle.id,
-                self.climate_config,
-            )
-            self.vehicle.air_control_is_on = True
+        try:
+            if hvac_mode == HVACMode.OFF:
+                await self.hass.async_add_executor_job(
+                    self.vehicle_manager.stop_climate,
+                    self.vehicle.id,
+                )
+                self.vehicle.air_control_is_on = False
+            else:
+                await self.hass.async_add_executor_job(
+                    self.vehicle_manager.start_climate,
+                    self.vehicle.id,
+                    self.climate_config,
+                )
+                self.vehicle.air_control_is_on = True
+        except UnsupportedControlError as ex:
+            _LOGGER.warning("Climate control not supported by vehicle: %s", ex)
         self.coordinator.async_request_refresh()
         self.async_write_ha_state()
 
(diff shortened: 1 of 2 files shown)


cdnninja:
We had a reason we did this. I will have to go back and look.

blka:
@cdnninja  did you find anything useful?

cdnninja:
I think it was removed during the 2.0 re-write and never readded.

custom_components/kia_uvo/climate.py
```diff
@@ -203,19 +203,22 @@ def supported_features(self) -> int:
...
+                    self.vehicle.id,
+                )
+                self.vehicle.air_control_is_on = False
+            else:
+                await self.hass.async_add_executor_job(
+                    self.vehicle_manager.start_climate,
+                    self.vehicle.id,
+                    self.climate_config,
+                )
+                self.vehicle.air_control_is_on = True
```
cdnninja: I am not sure we should internally do this in the kia_uvo side. I am open to optimistic setting in the library though.  Thoughts?

custom_components/kia_uvo/climate.py
```diff
@@ -203,19 +203,22 @@ def supported_features(self) -> int:
...
+        except UnsupportedControlError as ex:
+            _LOGGER.warning("Climate control not supported by vehicle: %s", ex)
```
cdnninja: Would this mean no toast is provided?

custom_components/kia_uvo/climate.py
blka: Re: optimistic state setting — good point. The `air_control_is_on = True/False` lines are legacy optimistic state from the original code. Since we already call `async_request_refresh()` + `async_write_ha_state()` right after, the refresh cycle will reconcile the real state from the API. I'll remove the optimistic lines.

Re: toast — yes, currently catching `UnsupportedControlError` and logging a warning means no toast/notification to the user. The alternative is to re-raise as `HomeAssistantError`, which would show a toast in HA. For vehicles that report climate state but can't control it remotely (like HEV), this would be the right UX — the user sees "not supported" instead of silent failure. I'll change the catch to re-raise as `HomeAssistantError` so the user gets feedback.

jwefers:
nice to see my original impl getting some love. I was wondering at the time why it was suddenly gone, but never had time to revisit it. Thanks.


NOTE: I noticed in the new myHuyndai app, that i can in fact change the target temp while hvac preconditioning is running, contrary to the code. I think this "stop, wait, start with new target" could be adapted. This is with my Ioniq electric facelift

blka:
> nice to see my original impl getting some love. I was wondering at the time why it was suddenly gone, but never had time to revisit it. Thanks.
> 
> 
> 
> 
> 
> NOTE: I noticed in the new myHuyndai app, that i can in fact change the target temp while hvac preconditioning is running, contrary to the code. I think this "stop, wait, start with new target" could be adapted. This is with my Ioniq electric facelife

Not sure about the historical stuff since I’ve only been messing with Hyundai for like two months, but I’m happy ’cause you’re happy 😏

Issue #1135, opened 2025-05-09
Climate Start does not work


FuzzyMistborn:
Please check Services, Known Bug / Issues and Troubleshooting over here first: https://github.com/fuatakgun/kia_uvo/blob/master/README.md
**Region and Brand of car**
USA, Hyundai Ioniq 5

**Version of the integration**
2.41.0

**Describe the bug**

I try to start the car via the climate start service and nothing happens.

**Debug logs if an error occurred**
```
2025-05-08 22:11:15.113 DEBUG (SyncWorker_10) [hyundai_kia_connect_api.HyundaiBlueLinkApiUSA] hyundai_kia_connect_api - Start engine..
[...]

PR #1645, opened 2026-05-07
feat: add CCS2 battery thermal and power consumption sensors

blka:
## Summary

Adds 8 new sensors for EV battery thermal monitoring, power consumption breakdown, and location staleness — data available via the CCS2 (DC fast charging) protocol.

### New sensors

| Key | Translation key | Unit | Device class |
|---|---|---|---|
| `ev_battery_pack_voltage` | `ev_battery_pack_voltage` | V | `VOLTAGE` |
| `ev_battery_temperature_min` | `ev_battery_temperature_min` | °C | `TEMPERATURE` |
| `ev_battery_temperature_max` | `ev_battery_temperature_max` | °C | `TEMPERATURE` |
| `ev_battery_water_temperature` | `ev_battery_water_temperature` | °C | `TEMPERATURE` |
| `ev_power_consumption_air_conditioning` | `ev_power_consumption_air_conditioning` | W | `POWER` |
| `ev_power_consumption_battery_cooling` | `ev_power_consumption_battery_cooling` | W | `POWER` |
| `ev_power_consumption_battery_heater` | `ev_power_consumption_battery_heater` | W | `POWER` |
| `location_last_updated_at` | `location_last_updated_at` | — | `TIMESTAMP` |

All new sensors use `getattr(vehicle, description.key, None) is not None` for conditional creation — they will only appear on vehicles that report these values.

### Testing

- All 8 sensors have values: `ev_battery_pack_voltage=259.3`, `ev_battery_temperature_min=24`, `ev_battery_temperature_max=25`, etc.

Depends on: #1641 (bug fixes)

## Test plan

- [ ] Install on vehicle with CCS2 support and verify 8 new sensor entities appear
- [ ] Verify voltage, temperature, and power readings are sensible
- [ ] Verify `location_last_updated_at` shows correct timestamp
- [ ] Install on vehicle without CCS2 data and verify no extra sensors are created


custom_components/kia_uvo/sensor.py
```diff
@@ -272,6 +274,76 @@
...
+        native_unit_of_measurement=UnitOfElectricPotential.VOLT,
+        entity_category=EntityCategory.DIAGNOSTIC,
+    ),
+    SensorEntityDescription(
+        key="ev_battery_temperature_min",
+        translation_key="ev_battery_temperature_min",
+        icon="mdi:thermometer-low",
+        device_class=SensorDeviceClass.TEMPERATURE,
+        state_class=SensorStateClass.MEASUREMENT,
+        native_unit_of_measurement=UnitOfTemperature.CELSIUS,
```
cdnninja: Could this come from the car instead of hard coding?  Same goes for others.

cdnninja (changes requested):

blka:
Responding to @cdnninja's review comment about hardcoded `UnitOfTemperature.CELSIUS` on battery temperature sensors:

I checked the Vehicle model — unlike `_air_temperature` and `_outside_temperature`, the battery temperature attributes (`ev_battery_temperature_min`, `ev_battery_temperature_max`, `ev_battery_water_temperature`) do **not** have corresponding `_unit` counterparts in the API response. The EU API always returns these values in °C with no unit indicator, so hardcoding `UnitOfTemperature.CELSIUS` is the correct approach here.

The sensors that use `DYNAMIC_UNIT` (like air temperature, outside temperature, driving range) all have `_unit` attributes in the Vehicle model that get populated from the API response. These battery temperature fields don't — the API simply provides integer values that are always Celsius.

cdnninja:
Doesn't the library have unit though? I do understand each region will have assumptions for this.  USA being F. Everyone else being C.

blka:
@cdnninja You're right — the library does have a unit pattern for temperature fields. I've updated both the API library and this PR to follow it properly:

**API library (PR #1135):** Battery temperature fields now use the same `_value`/`_unit` property pattern as `air_temperature` and `outside_temperature`. `ApiImplType1` sets them as `(value, TEMPERATURE_UNITS[0])` tuples (°C), since CCS2 BMS data is always reported in Celsius.

**This PR:** Changed the three battery temperature sensors from hardcoded `UnitOfTemperature.CELSIUS` to `DYNAMIC_UNIT`. The sensor class already has logic at line 414-415 that reads `_unit` from the vehicle when `native_unit_of_measurement == DYNAMIC_UNIT`:

```python
if self.entity_description.native_unit_of_measurement == DYNAMIC_UNIT:
    return getattr(self.vehicle, self._key + "_unit")
```

So `ev_battery_temperature_min` will read from `vehicle.ev_battery_temperature_min_unit`, etc.

Live test result on my Santa Fe (EU, CCS2):
```
ev_battery_temperature_min: 27  →  ev_battery_temperature_min_unit: °C
ev_battery_temperature_max: 31  →  ev_battery_temperature_max_unit: °C
ev_battery_water_temperature: 0 →  ev_battery_water_temperature_unit: °C
```

This is consistent with how all other temperature/range/duration sensors work in the integration.

cdnninja:
Conflicts on this one.

cdnninja:
Thanks for this!

PR #1642, opened 2026-05-07
feat: add cover platform for individual window control

blka:
## Summary

Adds a new `cover` platform with 4 individual window entities and ventilation support.

### Changes

- **`cover.py`** (NEW): 4 cover entities — `front_left_window`, `front_right_window`, `back_left_window`, `back_right_window`
  - Supported features: `OPEN | CLOSE | STOP | SET_POSITION`
  - Position mapping: 0=CLOSED, 1-49=VENTILATION, 50-100=OPEN
  - `exists_fn`: `vehicle.front_left_window_is_open is not None` (conditional — only creates on vehicles with window state)
- **`__init__.py`**: Add `Platform.COVER` to platform list
- **`coordinator.py`**: Add `async_open_all_windows()`, `async_close_all_windows()`, `async_vent_all_windows()` using `WindowRequestOptions` and `WINDOW_STATE`
- **`strings.json`** + **`translations/en.json`**: Translation keys for 4 window covers

### Testing

Validated against live Hyundai EU API (Santa Fe HEV):
- All 4 window cover entities would be created (`front_left_window_is_open=0`, etc.)

Depends on: #1641 (bug fixes)

## Test plan

- [ ] Install on vehicle with window state support and verify 4 cover entities appear
- [ ] Test open/close/stop/set_position operations
- [ ] Verify ventilation position (position 1-49) works correctly
- [ ] Install on vehicle without window state and verify no cover entities are created

Issue #1180, opened 2025-07-05
Authentication error - Australia Region


mattyvau:
**Region and Brand of car**

Kia EV in the Australia Region

**Version of the integration**
2.42.3. I think it's been on that version and working up until now.

**Describe the bug**
Integration started failing about 24 hours ago, with it displaying `Failed setup, will retry: Config Not Ready: Login Failed`.
Attempted to add a new hub with my credentials, this gave the same error.
The Kia Connect android app is still working, so the system must not be completely down.

**Debug logs if an error occurred**
```
2025-07-05 10:04:55.174 DEBUG (SyncWorker_5) [hyundai_kia_connect_api.KiaUvoApiAU] hyundai_kia_connect_api - Get Device ID request: {'ccsp-service-id': '8acb778a-b918-4a8d-8624-73a0beb64289', 'ccsp-application-id': '4ad4dcde-be23-48a8-bc1c-91b94f5c06f8', 'Stamp': 'FFeotQQ05kIVUmm+stN7LDV4WQeaChEfMkrNDk96P64QOPL9NN2YM7vTN0xxCfg=', 'Content-Type': 'application/json;charset=UTF-8', 'Host': 'au-apigw.ccs.kia.com.au:8082', 'Connection': 'Keep-Alive', 'Accept-Encoding': 'gzip', 'User-Agent': 'okhttp/3.12.0'} {'pushRegId': '315e6633c73b18b05fda92bca5c859aecab415875e5a063f1c1b89c08ab8863a', 'pushType': 'GCM', 'uuid': '6577bc8b-c860-48fc-883a-45cdc5f648ff'}

2025-07-05 10:04:55.613 DEBUG (SyncWorker_5) [hyundai_kia_connect_api.KiaUvoApiAU] hyundai_kia_connect_api - Get Device ID response: {'retCode': 'S', 'resCode': '0000', 'resMsg': {'deviceId': 'f177740a-3e0e-4766-9607-10c653bc5588'}, 'msgId': 'ae8f81d0-5933-11f0-b255-0dcfc5dff21b'}

[...]

Issue #1167, opened 2025-06-24
Values not updating instantly


vineethktpla:
**KIA Sportage - EU 2022 HEV**

**Version of the integration- 2.41.2**

**Describe the bug**
The values are not updating instantly, for example If I unlock the car and lock it back, the UI still shows unlocked.
If the engine is running, it still shows engine not running. I am tracking my trips based on the engine running status, so that this method is not reliable. It takes a while for odometer and range, battery soc to update, which are not instantly updating after I stop the car.

cdnninja:
It isn't instant. This is normal behaviour.  It has a polling system.

PR #1647, opened 2026-05-07
feat: add action buttons for charge port, hazard, valet, and windows

blka:
## Summary

Adds 9 new action buttons for vehicle control operations.

### New buttons

| Key | Translation key | Type | Condition |
|---|---|---|---|
| `open_charge_port` | `open_charge_port` | Conditional | `vehicle.ev_charge_port_door_is_open is not None` |
| `close_charge_port` | `close_charge_port` | Conditional | `vehicle.ev_charge_port_door_is_open is not None` |
| `start_hazard_lights` | `start_hazard_lights` | Fire-and-forget | Always created |
| `start_hazard_lights_and_horn` | `start_hazard_lights_and_horn` | Fire-and-forget | Always created |
| `start_valet_mode` | `start_valet_mode` | Fire-and-forget | Always created |
| `stop_valet_mode` | `stop_valet_mode` | Fire-and-forget | Always created |
| `open_all_windows` | `open_all_windows` | Conditional | `vehicle.front_left_window_is_open is not None` |
| `close_all_windows` | `close_all_windows` | Conditional | `vehicle.front_left_window_is_open is not None` |
| `vent_all_windows` | `vent_all_windows` | Conditional | `vehicle.front_left_window_is_open is not None` |

Fire-and-forget buttons (hazard, valet) are always created because they have no readable state. Conditional buttons only appear when the vehicle supports the related feature.

### Changes

- **`button.py`**: Add 9 button descriptions with `press_action` mapping to coordinator methods
- **`coordinator.py`**: Add `async_open_all_windows()`, `async_close_all_windows()`, `async_vent_all_windows()` using `WindowRequestOptions` and `WINDOW_STATE`
- **`strings.json`** + **`translations/en.json`**: Translation keys for 9 buttons

### Testing

- 8/10 entities would be created (charge port buttons not created on HEV — `ev_charge_port_door_is_open is None`)
- Window button conditions met (`front_left_window_is_open=0`)
- All VehicleManager methods verified: `open_charge_port`, `close_charge_port`, `start_hazard_lights`, `start_hazard_lights_and_horn`, `start_valet_mode`, `stop_valet_mode`, `set_windows_state`

Depends on: #1641 (bug fixes), #1642 (cover — window coordinator methods)

## Test plan

- [ ] Install on EV vehicle and verify charge port buttons appear
- [ ] Install on vehicle with window state and verify window buttons appear
- [ ] Test hazard lights and valet mode buttons
- [ ] Verify fire-and-forget buttons don't block on action status polling
- [ ] Install on HEV vehicle and verify charge port buttons don't appear

### DIFF
--- a/custom_components/kia_uvo/button.py
+++ b/custom_components/kia_uvo/button.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from dataclasses import dataclass
 import logging
 from typing import Final
@@ -23,6 +24,8 @@ _LOGGER = logging.getLogger(__name__)
 @dataclass(frozen=True, kw_only=True)
 class HyundaiKiaButtonDescription(ButtonEntityDescription):
     press_action: str
+    exists_fn: Callable[[Vehicle], bool] = lambda _: True
+    enabled_fn: Callable[[Vehicle], bool] = lambda _: True
 
 
 BUTTON_DESCRIPTIONS: Final[tuple[HyundaiKiaButtonDescription, ...]] = (
@@ -31,6 +34,41 @@ BUTTON_DESCRIPTIONS: Final[tuple[HyundaiKiaButtonDescription, ...]] = (
         translation_key="force_refresh",
         icon="mdi:refresh",
         press_action="async_force_refresh_vehicle",
+    ),
+    HyundaiKiaButtonDescription(
+        key="start_hazard_lights",
+        translation_key="start_hazard_lights",
+        icon="mdi:hazard-lights",
+        press_action="async_start_hazard_lights",
+        enabled_fn=lambda _: False,
+    ),
+    HyundaiKiaButtonDescription(
+        key="start_hazard_lights_and_horn",
+        translation_key="start_hazard_lights_and_horn",
+        icon="mdi:car-emergency",
+        press_action="async_start_hazard_lights_and_horn",
+        enabled_fn=lambda _: False,
+    ),
+    HyundaiKiaButtonDescription(
+        key="open_all_windows",
+        translation_key="open_all_windows",
+        icon="mdi:window-maximize",
+        press_action="async_open_all_windows",
+        exists_fn=lambda vehicle: vehicle.front_left_window_is_open is not None,
+    ),
+    HyundaiKiaButtonDescription(
+        key="close_all_windows",
+        translation_key="close_all_windows",
+        icon="mdi:window-minimize",
+        press_action="async_close_all_windows",
+        exists_fn=lambda vehicle: vehicle.front_left_window_is_open is not None,
+    ),
+    HyundaiKiaButtonDescription(
+        key="vent_all_windows",
+        translation_key="vent_all_windows",
+        icon="mdi:window-open-variant",
+        press_action="async_vent_all_windows",
+        exists_fn=lambda vehicle: vehicle.front_left_window_is_open is not None,
     ),
 )
 
@@ -45,7 +83,10 @@ async def async_setup_entry(
     for vehicle_id in coordinator.vehicle_manager.vehicles.keys():
         vehicle: Vehicle = coordinator.vehicle_manager.vehicles[vehicle_id]
         for description in BUTTON_DESCRIPTIONS:
-            entities.append(HyundaiKiaConnectButton(coordinator, description, vehicle))
+            if description.exists_fn(vehicle):
+                entities.append(
+                    HyundaiKiaConnectButton(coordinator, description, vehicle)
+                )
 
     async_add_entities(entities)
 
@@ -65,6 +106,7 @@ class HyundaiKiaConnectButton(ButtonEntity, HyundaiKiaConnectEntity):
         self._key = description.key
         self._attr_unique_id = f"{DOMAIN}_{vehicle.id}_{self._key}"
         self._attr_icon = description.icon
+        self._attr_entity_registry_enabled_default = description.enabled_fn(vehicle)
 
     async def async_press(self) -> None:
         await getattr(self.coordinator, self.entity_description.press_action)(
--- a/custom_components/kia_uvo/button.py
+++ b/custom_components/kia_uvo/button.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from dataclasses import dataclass
 import logging
 from typing import Final
@@ -23,6 +24,7 @@ _LOGGER = logging.getLogger(__name__)
 @dataclass(frozen=True, kw_only=True)
 class HyundaiKiaButtonDescription(ButtonEntityDescription):
     press_action: str
+    exists_fn: Callable[[Vehicle], bool] = lambda _: True
 
 
 BUTTON_DESCRIPTIONS: Final[tuple[HyundaiKiaButtonDescription, ...]] = (
@@ -33,58 +35,39 @@ BUTTON_DESCRIPTIONS: Final[tuple[HyundaiKiaButtonDescription, ...]] = (
         press_action="async_force_refresh_vehicle",
     ),
     HyundaiKiaButtonDescription(
-        key="open_charge_port",
-        translation_key="open_charge_port",
-        icon="mdi:ev-plug-charging",
-        press_action="async_open_charge_port",
-    ),
-    HyundaiKiaButtonDescription(
-        key="close_charge_port",
-        translation_key="close_charge_port",
-        icon="mdi:ev-plug",
-        press_action="async_close_charge_port",
-    ),
-    HyundaiKiaButtonDescription(
         key="start_hazard_lights",
         translation_key="start_hazard_lights",
         icon="mdi:hazard-lights",
         press_action="async_start_hazard_lights",
+        exists_fn=lambda vehicle: vehicle.supports_window_control is not None,
     ),
     HyundaiKiaButtonDescription(
         key="start_hazard_lights_and_horn",
         translation_key="start_hazard_lights_and_horn",
         icon="mdi:car-emergency",
         press_action="async_start_hazard_lights_and_horn",
-    ),
-    HyundaiKiaButtonDescription(
-        key="start_valet_mode",
-        translation_key="start_valet_mode",
-        icon="mdi:key-variant",
-        press_action="async_start_valet_mode",
-    ),
-    HyundaiKiaButtonDescription(
-        key="stop_valet_mode",
-        translation_key="stop_valet_mode",
-        icon="mdi:key",
-        press_action="async_stop_valet_mode",
+        exists_fn=lambda vehicle: vehicle.supports_window_control is not None,
     ),
     HyundaiKiaButtonDescription(
         key="open_all_windows",
         translation_key="open_all_windows",
         icon="mdi:window-maximize",
         press_action="async_open_all_windows",
+        exists_fn=lambda vehicle: vehicle.front_left_window_is_open is not None,
     ),
     HyundaiKiaButtonDescription(
         key="close_all_windows",
         translation_key="close_all_windows",
         icon="mdi:window-minimize",
         press_action="async_close_all_windows",
+        exists_fn=lambda vehicle: vehicle.front_left_window_is_open is not None,
     ),
     HyundaiKiaButtonDescription(
         key="vent_all_windows",
         translation_key="vent_all_windows",
         icon="mdi:window-open-variant",
         press_action="async_vent_all_windows",
+        exists_fn=lambda vehicle: vehicle.front_left_window_is_open is not None,
     ),
 )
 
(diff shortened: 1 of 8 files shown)


cdnninja:
I thought windows was added as a cover entity?  

Looks like some of these should be switches?   Or would it be too slow from a response standpoint for things like hazard?

custom_components/kia_uvo/coordinator.py
```diff
@@ -504,6 +504,61 @@ async def async_set_windows(
...
+            back_left=WINDOW_STATE.VENTILATION,
+            back_right=WINDOW_STATE.VENTILATION,
+        )
+        await self._async_send_action(
+            vehicle_id,
+            lambda: self.vehicle_manager.set_windows_state(vehicle_id, options),
+            "vent all windows",
+        )
+
+    async def _async_send_action(self, vehicle_id: str, action_fn, action_name: str):
```
cdnninja: I think we have something similar that I just merged.

custom_components/kia_uvo/coordinator.py
```diff
@@ -504,6 +504,61 @@ async def async_set_windows(
...
+            back_right=WINDOW_STATE.CLOSED,
+        )
+        await self._async_send_action(
+            vehicle_id,
+            lambda: self.vehicle_manager.set_windows_state(vehicle_id, options),
+            "close all windows",
+        )
+
+    async def async_vent_all_windows(self, vehicle_id: str):
+        from hyundai_kia_connect_api.const import WINDOW_STATE
```
cdnninja: No imports inline.

custom_components/kia_uvo/coordinator.py
```diff
@@ -504,6 +504,61 @@ async def async_set_windows(
...
+            back_left=WINDOW_STATE.OPEN,
+            back_right=WINDOW_STATE.OPEN,
+        )
+        await self._async_send_action(
+            vehicle_id,
+            lambda: self.vehicle_manager.set_windows_state(vehicle_id, options),
+            "open all windows",
+        )
+
+    async def async_close_all_windows(self, vehicle_id: str):
```
cdnninja, line 553: Is this worthwhile vs the existing window command?

cdnninja (changes requested):

custom_components/kia_uvo/coordinator.py
blka: Good catch — `_async_send_action` was already merged in #1635. I've rebased onto master and removed the duplicate. The branch now only adds the window convenience methods (`open_all_windows`, `close_all_windows`, `vent_all_windows`) which delegate to the existing `_async_send_action` from #1635.

custom_components/kia_uvo/coordinator.py
blka: Fixed — moved `WINDOW_STATE` to top-level imports, no more inline imports.

custom_components/kia_uvo/coordinator.py
blka: Good question. The existing `set_windows` service (via cover entity) controls individual windows with granular state. These button convenience methods are for the common case of "open all / close all / vent all" — one tap instead of adjusting each window cover individually. 

The cover entity is the right UX for per-window control. The buttons are for quick bulk actions. Both use the same underlying `set_windows_state` API call, just with different `WindowRequestOptions`.

If you'd prefer these as switches instead of buttons (since they have a toggling feel for open/close), I can convert. But hazard lights and valet mode are fire-and-forget with no toggle state — buttons are the right entity type for those.

blka:
Rebased onto master. Changes since last review:

1. **Removed duplicate `_async_send_action`** — #1635 already merged this with action lock + `UnsupportedControlError` handling. This PR now only adds the net-new window convenience methods, which delegate to the existing helper.
2. **Moved `WINDOW_STATE` import to top-level** — no more inline imports per review feedback.

Regarding the window buttons vs cover entity question: the cover entity (`set_windows` service) is for per-window granular control. The `open_all_windows` / `close_all_windows` / `vent_all_windows` buttons are convenience shortcuts for the common "open everything / close everything" use case — one tap instead of adjusting 4 cover entities. Same underlying API call, different UX intent.

For hazard lights, valet mode, and charge port — these are fire-and-forget actions with no toggle state, so buttons are the correct entity type. Switches would imply a state that can be read back, which these don't have.

Marking as ready for review.

blka:
@cdnninja if we are ok with all the fixes, could you take care or this PR?

custom_components/kia_uvo/coordinator.py
blka: ?

custom_components/kia_uvo/button.py
```diff
@@ -45,6 +99,16 @@ async def async_setup_entry(
     for vehicle_id in coordinator.vehicle_manager.vehicles.keys():
         vehicle: Vehicle = coordinator.vehicle_manager.vehicles[vehicle_id]
         for description in BUTTON_DESCRIPTIONS:
+            if description.key in ("open_charge_port", "close_charge_port"):
```
cdnninja: Could we add a exists_fn to the description so we don't need unique items for each one?

custom_components/kia_uvo/button.py
```diff
@@ -32,6 +32,60 @@ class HyundaiKiaButtonDescription(ButtonEntityDescription):
...
+        press_action="async_start_hazard_lights",
+    ),
+    HyundaiKiaButtonDescription(
+        key="start_hazard_lights_and_horn",
+        translation_key="start_hazard_lights_and_horn",
+        icon="mdi:car-emergency",
+        press_action="async_start_hazard_lights_and_horn",
+    ),
+    HyundaiKiaButtonDescription(
+        key="start_valet_mode",
```
cdnninja: Should this be a switch?

custom_components/kia_uvo/button.py
```diff
@@ -32,6 +32,60 @@ class HyundaiKiaButtonDescription(ButtonEntityDescription):
         icon="mdi:refresh",
         press_action="async_force_refresh_vehicle",
     ),
+    HyundaiKiaButtonDescription(
+        key="open_charge_port",
```
cdnninja: Should this be a switch?

custom_components/kia_uvo/coordinator.py
cdnninja: Whenever on off exists with state it should not be a button.

cdnninja (changes requested):
A few more comments. I am a bit worried with some of these items adding clutter or adding them for those that can't use them due to model of car.    How do we make sure this only shows for people who can use it?

custom_components/kia_uvo/coordinator.py
blka: Hi @cdnninja, I've addressed all your feedback from the latest review:

1. **`exists_fn` on button descriptions** — Added `exists_fn` to `HyundaiKiaButtonDescription` so entities only show for vehicles that support them. Hazard lights use `vehicle.supports_window_control is not None`, windows use `vehicle.front_left_window_is_open is not None`. No more hardcoded `if` blocks in `async_setup_entry`.

2. **Charge port → switch** — Moved from button to switch. `ev_charge_port_door_is_open` has clear on/off state, so a switch is the correct entity type. `on_fn` = open, `off_fn` = close.

3. **Valet mode → switch** — Moved from button to switch. `valet_mode_active` has clear on/off state. `on_fn` = activate, `off_fn` = deactivate.

4. **Windows stay as buttons** — Windows have 3 states (closed/vent/open), so they don't map cleanly to binary on/off. `vent_all_windows` is also incremental — each press vents a bit more. Buttons are the right fit.

5. **Hazard lights stay as buttons** — Fire-and-forget action with no readable state. No on/off to toggle.

6. **Clutter concern** — The `exists_fn` pattern ensures entities only appear for vehicles that actually support each feature. Charge port switch appears only if `ev_charge_port_door_is_open is not None`, valet switch only if `valet_mode_active is not None`, etc.

Branch has been rebased onto latest master (conflicts resolved with the new charging schedule switches from #1710).

Ready for re-review.

blka:
@cdnninja — rebased onto latest master (clean, 3 commits) and fixed one issue I spotted while re-reviewing my own diff:

The refactor commit had accidentally pasted the window-control `exists_fn` onto the hazard/horn button descriptions:

```python
exists_fn=lambda vehicle: vehicle.supports_window_control is not None,
```

Hazard/horn are fire-and-forget with no readable state — they should always be created (default `exists_fn = lambda _: True`), matching the PR description and the refactor commit's own message ("Hazard lights: keep as buttons (fire-and-forget, no state)"). The condition was functionally always-True anyway (every region sets `supports_window_control` to True/False, never None), but it was misleading. Dropped it so hazard/horn use the default.

`exists_fn` is now only on the genuinely conditional buttons (open/close/vent windows → `front_left_window_is_open is not None`). This matches your original ask — exists_fn for conditional buttons, default for always-created fire-and-forget ones.

Separately: the underlying API methods for horn/hazard USA just landed upstream in hyundai_kia_connect_api (#1180 + #1191 — the polling fix was POST→GET). Live-tested by a USA user (2025 Tucson Hybrid) in API issue #1167 — commands work, polling clean. So once this kia_uvo PR merges, USA users get working hazard/horn buttons on the device page.

Ready for re-review when you have a moment.

blka:
@cdnninja Re-review request — I believe all the requested changes are addressed in the latest commits on this branch. Mapping your comments to what changed:

- **`coordinator.py:552` ("something similar that I just merged")** — this was the `exists_fn`/device-page button pattern; folded into the button descriptions (see below).
- **`coordinator.py:538` ("no imports inline")** — inlined import removed; coordinator keeps imports at the top.
- **`coordinator.py:553` ("worthwhile vs the existing window command?")** — kept the window buttons, but they reuse the existing window command path rather than a parallel one.
- **`button.py:102` ("add an `exists_fn` to the description so we don't need unique items for each one")** — added `exists_fn` to the button descriptions.
- **`button.py:60` ("Should this be a switch?")** — charge port moved to a switch.
- **`button.py:36` ("Should this be a switch?")** — valet moved to a switch.

Net: charge port + valet are now switches (toggle), hazard + horn + windows stay as buttons on the device page. Commit `4dbf994` (move charge port and valet to switches, add exists_fn) and `6184e85` (drop bogus exists_fn from hazard/horn buttons) cover it.

Could you take another look when you get a chance? @ijojog is also blocked on this — once it merges, the hazard/horn buttons become available in stable `kia_uvo` (the API-side horn/hazard from #1180 is already merged; this PR is the HA-side wiring). Thanks.

blka:
Housekeeping: rebased onto current master and squashed the three commits (feat + refactor + exists_fn fix) into a single commit `b0a0c36`. The review-item mapping in my previous comment still holds — charge port + valet are switches, hazard/horn/windows are buttons with exists_fn only on the conditional window buttons. PR is mergeable (1 commit, clean against master). The commit hashes referenced earlier (`4dbf994`, `6184e85`) no longer exist after the squash.

cdnninja:
I think my worry is too many items that not everyone will use that they can add themselves.  As well I suspect not all of these will work for all car types yet they will display.   

Thoughts?  Could we add in a disabled state?

ijojog:
That makes sense. Just sharing my perspective as a user/tester:

I already see some entities/services in my setup that do not apply to my vehicle, for example EV/charging-related items, and personally that does not bother me. I think most users understand that not every vehicle supports every feature, and they can ignore or disable the entities they do not use.

That said, I also understand the concern about clutter and features showing up for vehicles where they may not work.

Maybe a good middle ground would be:

```text
1. Add the entities, but default some of them to disabled if support cannot be confidently detected.
2. Enable/display automatically only when there is a reliable capability flag or exists_fn.
3. Let users manually enable the disabled entities if they know their vehicle supports the feature.
```

For example:

```text
- Hazard lights / horn could be available or disabled-by-default depending on region/support confidence.
- Charge port / valet as switches makes sense.
- Window buttons could stay conditional if exists_fn can detect support.
```

Another possible option would be grouping/filtering by vehicle type:

```text
ICE / HEV / PHEV / EV
```

But I would be careful with vehicle-type-only filtering, because some HEV/PHEV/EV feature overlap is not always obvious. Some vehicles may support features that another vehicle of the same broad type does not.

So my preference as a user would be:

```text
Use capability detection where possible.
If uncertain, create the entity but disabled by default.
Let advanced users enable it manually.
```

That way the integration does not become too cluttered for everyone, but the functionality is still available for users whose vehicles support it.

blka:
@cdnninja @ijojog — good call on the disabled-by-default concern. I pushed a change that addresses it while keeping the useful buttons available.

What changed:

- start_hazard_lights and start_hazard_lights_and_horn are now created but disabled by default (entity_registry_enabled_default=False).
  - Reason: we have no reliable vehicle/region capability flag for horn/hazard, so we cannot confidently auto-enable them.
  - Users who know their vehicle supports it can enable the buttons manually in the entity registry.

- open_all_windows, close_all_windows, and vent_all_windows stay enabled by default.
  - Reason: they already have a solid exists_fn based on reported window state (front_left_window_is_open is not None).

- Charge-port and valet-mode switches remain enabled by default because their existence is gated by readable state (ev_charge_port_door_is_open is not None, valet_mode_active is not None).

Implementation:

- Added enabled_fn to HyundaiKiaButtonDescription and wired it to self._attr_entity_registry_enabled_default in the button entity __init__.
- Only hazard/horn set enabled_fn=lambda _: False; everything else keeps the default True.

Tests:

- Added tests/test_button_entity_descriptions.py covering exists_fn and enabled_fn predicates.
- Added tests/conftest.py and pytest.ini with asyncio_mode=auto so the tests can run alongside pytest-homeassistant-custom-component autouse async fixtures.
- Branch rebased onto current upstream/master and squashed to one commit: 78b3d48.

@cdnninja this should resolve the clutter concern while still letting users opt into the new actions. Ready for re-review when you have a moment.

ijojog:
That sounds like a good compromise to me.

As a user/tester, I like this approach:

```text
- hazard/horn available but disabled by default
- users who know their vehicle supports them can manually enable them
- features with reliable capability/state detection can stay enabled by default
```

That should avoid clutter for most users while still making the working USA horn/hazard actions available.

I’m happy to test once this is merged or if you need me to test the branch directly.

PR #1649, opened 2026-05-07
feat: add EV thermal diagnostics and outside temperature

blka:
## Summary

Adds sensors for EV battery thermal management, outside temperature, and engine type identification.

### New sensors (3)

| Key | Translation key | Unit | Device class |
|---|---|---|---|
| `outside_temperature` | `outside_temperature` | Dynamic (°C/°F from API) | `TEMPERATURE` |
| `engine_type` | `engine_type` | — | Text sensor (EV/PHEV/HEV/IC) |
| `ev_battery_chiller_rpm` | `ev_battery_chiller_rpm` | RPM | — |

The `outside_temperature` sensor uses `DYNAMIC_UNIT` with the `_key` prefix convention (`_outside_temperature`), so HA resolves the unit from `Vehicle._outside_temperature_unit` at runtime — ensuring correct °F display for US users.

### New binary sensors (2)

| Key | Translation key |
|---|---|
| `ev_battery_precondition_enabled` | `ev_battery_precondition_enabled` |
| `ev_battery_heating_state` | `ev_battery_heating_state` |

All new entities use `getattr(vehicle, description.key, None) is not None` for conditional creation.

### Testing

Validated against live Hyundai EU API (Santa Fe HEV):
- `engine_type=ENGINE_TYPES.HEV` — engine type sensor works correctly
- Other thermal diagnostics (`ev_battery_chiller_rpm`, `ev_battery_precondition_enabled`, `ev_battery_heating_state`) are `None` on this HEV — correctly would not be created
- `outside_temperature` is `None` on this vehicle

Depends on: #1645 (CCS2 sensors — `ev_battery_chiller_rpm` is a thermal sensor)

## Test plan

- [ ] Install on EV vehicle and verify thermal diagnostic entities appear
- [ ] Verify `engine_type` shows correct type string (EV/PHEV/HEV/IC)
- [ ] Verify `outside_temperature` shows correct value with proper unit (°C or °F)
- [ ] Install on HEV vehicle and verify only `engine_type` entity appears

PR #1648, opened 2026-05-07
feat: add EV departure schedule entities

blka:
## Summary

Adds 8 new entities completing the EV departure schedule: days, climate settings, and defrost.

### New sensors (4)

| Key | Translation key | Type | Note |
|---|---|---|---|
| `ev_first_departure_days` | `ev_first_departure_days` | Text sensor | Comma-separated day names (e.g., "Mo,Tu,We") |
| `ev_second_departure_days` | `ev_second_departure_days` | Text sensor | Comma-separated day names |
| `ev_first_departure_climate_temperature` | `ev_first_departure_climate_temperature` | Temperature sensor | Dynamic unit (°C/°F from API) |
| `ev_second_departure_climate_temperature` | `ev_second_departure_climate_temperature` | Temperature sensor | Dynamic unit (°C/°F from API) |

The departure days attributes are lists in the Vehicle model. They are rendered as comma-separated text strings in the sensor (read-only from HA perspective).

Departure climate temperature sensors use `DYNAMIC_UNIT` with `_key` prefix convention, so HA resolves the unit from `Vehicle._ev_first_departure_climate_temperature_unit` / `_ev_second_departure_climate_temperature_unit` at runtime — ensuring correct °F display for US users.

### New binary sensors (4)

| Key | Translation key | Note |
|---|---|---|
| `ev_first_departure_climate_enabled` | `ev_first_departure_climate_enabled` | |
| `ev_second_departure_climate_enabled` | `ev_second_departure_climate_enabled` | |
| `ev_first_departure_climate_defrost` | `ev_first_departure_climate_defrost` | |
| `ev_second_departure_climate_defrost` | `ev_second_departure_climate_defrost` | |

All new entities use `getattr(vehicle, description.key, None) is not None` for conditional creation.

### Testing

Validated against live Hyundai EU API (Santa Fe HEV):
- Departure schedule fields are `None` on HEV — entities correctly would not be created
- On EV/PHEV vehicles with departure schedules configured, these entities would appear

Depends on: #1641 (bug fixes)

## Test plan

- [ ] Install on EV vehicle with departure schedules and verify all 8 entities appear
- [ ] Verify departure days display as comma-separated strings
- [ ] Verify climate temperature sensors show correct values with proper unit (°C or °F)
- [ ] Verify binary sensors for climate enabled/defrost show correct on/off state
- [ ] Install on HEV vehicle and verify no departure entities are created

PR #1650, opened 2026-05-07
feat: add V2L/V2X status and charging schedule binary sensors

blka:
## Summary

Adds 4 new binary sensors for Vehicle-to-Load (V2L) / Vehicle-to-Everything (V2X) status and charging schedule controls.

### New binary sensors

| Key | Translation key | Description |
|---|---|---|
| `ev_v2l_status` | `ev_v2l_status` | V2L active status |
| `ev_v2x_status` | `ev_v2x_status` | V2X active status |
| `ev_off_peak_charge_only_enabled` | `ev_off_peak_charge_only_enabled` | Off-peak charge only schedule enabled |
| `ev_schedule_charge_enabled` | `ev_schedule_charge_enabled` | Scheduled charging enabled |

All new entities use `getattr(vehicle, description.key, None) is not None` for conditional creation — they will only appear on vehicles that report these values.

### Testing

- All V2L/V2X and charging schedule attributes are `None` on this HEV — entities correctly would not be created
- On EV/PHEV vehicles with V2L support, these entities would appear

Depends on: #1641 (bug fixes)

## Test plan

- [ ] Install on EV vehicle with V2L/V2X support and verify binary sensors appear
- [ ] Verify V2L/V2X status correctly reflects on/off state
- [ ] Verify charging schedule enabled/disabled state
- [ ] Install on HEV vehicle and verify no extra entities are created

### DIFF
--- a/custom_components/kia_uvo/binary_sensor.py
+++ b/custom_components/kia_uvo/binary_sensor.py
@@ -494,6 +494,18 @@ SENSOR_DESCRIPTIONS: Final[tuple[HyundaiKiaBinarySensorEntityDescription, ...]] 
         is_on=lambda vehicle: vehicle.ev_battery_heating_state,
         entity_category=EntityCategory.DIAGNOSTIC,
     ),
+    HyundaiKiaBinarySensorEntityDescription(
+        key="ev_v2l_status",
+        translation_key="ev_v2l_status",
+        icon="mdi:ev-station",
+        is_on=lambda vehicle: vehicle.ev_v2l_status,
+    ),
+    HyundaiKiaBinarySensorEntityDescription(
+        key="ev_v2x_status",
+        translation_key="ev_v2x_status",
+        icon="mdi:ev-station",
+        is_on=lambda vehicle: vehicle.ev_v2x_status,
+    ),
 )
 
 
--- a/custom_components/kia_uvo/strings.json
+++ b/custom_components/kia_uvo/strings.json
@@ -243,6 +243,12 @@
       },
       "location_last_updated_at": {
         "name": "Location Last Updated"
+      },
+      "ev_v2l_max_power": {
+        "name": "V2L max power"
+      },
+      "ev_v2l_target_soc": {
+        "name": "V2L target SoC"
       }
     },
     "binary_sensor": {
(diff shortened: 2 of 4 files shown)


cdnninja:
Should some of these be switches not binary sensors? 

It also has merge conflicts.

blka:
> Should some of these be switches not binary sensors?

i will replace it with switches, but need someone with PHEV to test it, cause some extra logic around will change. do you know anyone?

blka:
Good catch — you're right that some of these should be switches, not binary sensors.

I've converted the toggleable ones to switches:

**Converted to switches (can be toggled via `schedule_charging_and_climate` API):**
- `ev_schedule_charge_enabled` → switch
- `ev_off_peak_charge_only_enabled` → switch

**Staying as binary sensors (read-only status):**
- `ev_v2l_status` — read-only status (V2L cable connected/active)
- `ev_v2x_status` — read-only status (V2X active)

Each switch reads the current vehicle state via `_build_schedule_options_from_vehicle()`, changes only its specific field, and calls `schedule_charging_and_climate` with the full state preserved. This avoids accidentally resetting other schedule settings when toggling one.

I've also done the same conversion in PR #1648 (departure schedule) for `ev_first/second_departure_enabled`, `ev_first/second_departure_climate_enabled`, and `ev_first/second_departure_climate_defrost`.

For `ev_battery_precondition_enabled` in PR #1649 — there's no dedicated API method to toggle battery preconditioning yet, so I'm keeping it as a binary sensor for now. I'll open a follow-up issue to convert it once the API library adds support.

cdnninja:
Conflicts on this one currently.

cdnninja:
This has merge conflicts.   Ideally each PR doesn't depend or repeat items, it will make merging cleaner.

cdnninja:
Please mark as ready for review when good to go.

blka:
Rebased onto latest master — all merge conflicts resolved. Changes:

1. **Removed duplicate `_async_send_action`** — now uses the one from merged #1635 with action lock and UnsupportedControlError handling
2. **Converted toggleable entities to switches** per cdnninja's feedback:
   - `ev_schedule_charge_enabled` → switch (can be toggled via `schedule_charging_and_climate` API)
   - `ev_off_peak_charge_only_enabled` → switch (can be toggled)
   - V2L/V2X status remain as binary sensors (read-only)
3. **`_build_schedule_options_from_vehicle()`** — new helper that reads current vehicle state and builds full `ScheduleChargingClimateRequestOptions`, so each switch only changes its specific field without resetting other schedule settings
4. **Merged with PR #1648** — departure schedule switches are included in the diff now since both are on master; no duplicate entities

No overlap with other PRs anymore — each PR only adds net-new entities.

blka:
@cdnninja if we are ok with all the fixes, could you take care or this PR?

PR #1677, opened 2026-05-23
fix: reduce duplicate calls

cdnninja:
Certain calls were calling a refresh after they refreshed. Forced update calls should be parsing data themselves in all regions so not calling cached after since redundant.

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -185,7 +185,7 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
         await self.hass.async_add_executor_job(
             self.vehicle_manager.update_all_vehicles_with_cached_state
         )
-        await self.async_refresh()
+        self.async_set_updated_data(self.data)
 
     async def async_force_update_all(self) -> None:
         """Force refresh vehicle data and update it."""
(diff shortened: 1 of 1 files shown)


cdnninja:
@blka  Are you able to take a look and review?  This could have been related also to duplicate calls.

blka (approved):
Approve — the core change is correct and consistent with the existing pattern.

`async_set_updated_data(self.data)` is the right call here. The executor job (`update_all_vehicles_with_cached_state` / `force_refresh_all_vehicles_states`) already populates the vehicle objects in-place, so a full `async_refresh()` just re-runs `_async_update_data` and makes a second identical API call. This was already the pattern used in `async_force_refresh_vehicle` (line 196) and `async_await_action_and_force_refresh` (line 246), so this PR makes the three methods consistent.

This should also reduce `DuplicateRequestError` hits in EU, since `force_refresh_all_vehicles_states` internally sends a remote command and then polls — the extra `async_refresh()` on top was firing another cached-state call that could overlap with the still-in-flight action.

**One thing to note for a follow-up:** `async_await_action_and_refresh` (line 216) still calls `await self.async_refresh()` after action polling. This is used by lock/unlock/climate/start-charge/stop-charge/etc. That `async_refresh()` will trigger another `update_all_vehicles_with_cached_state` via `_async_update_data`. In most cases this is redundant since the action status polling already has an up-to-date vehicle state. Changing that to `self.async_set_updated_data(self.data)` would eliminate another class of duplicate calls, but it's a slightly different case (the action might have side effects that a cached poll wouldn't reflect), so leaving it for a separate PR makes sense.

Thanks for the ping on this — clean fix.

PR #1697, opened 2026-06-03
feat: add DayTripInfoEntity for per-trip sensor data

janfrederik:
## Summary

Adds a new `DayTripInfoEntity` sensor exposing per-trip data for today, plus the coordinator wiring to populate it. The library (`hyundai_kia_connect_api`) already implements `update_day_trip_info()`, but the upstream coordinator never calls it, so the data is unreachable from HA. This PR closes that gap.

## What it adds

A single new sensor per vehicle (translation key `day_trip_info`, e.g. `sensor.<car>_day_trip_info`):

- **State** — integer trip count for today (`0` if no trips yet or endpoint unsupported, never `unavailable`).
- **Attributes**:
  - `date` — ISO `YYYY-MM-DD` for the day represented.
  - `summary` — day totals object: `drive_time`, `idle_time`, `distance`, `avg_speed`, `max_speed` (or `null` before the first successful fetch).
  - `trip_list` — list of per-trip dicts with `start_time` and `end_time` (naive ISO 8601 local), `drive_time`, `idle_time`, `distance`, `avg_speed`, `max_speed`.

## How it works

`coordinator.py` — in the cached-state branch of `_async_update_data()`, after the existing `update_all_vehicles_with_cached_state()` call:

- Compute `today_str = dt_util.now().strftime("%Y%m%d")` once.
- Loop over `self.vehicle_manager.vehicles.items()` and call `vehicle_manager.update_day_trip_info(vehicle_id, today_str)` for EV/PHEV vehicles only.
- The library call is synchronous → wrapped in `async_add_executor_job`.
- The full call is wrapped in `try/except Exception` with a DEBUG log so a `NoDataFound`, an unsupported region/firmware, or any transient backend error becomes a debug line, **not** an `UpdateFailed`. This protects integration availability — the per-trip endpoint is reportedly absent or flaky in some regions.

`sensor.py` — adds `DayTripInfoEntity(SensorEntity, HyundaiKiaConnectEntity)`:

- Registered unconditionally in `async_setup_entry` (one per vehicle). The entity is created at startup even when `vehicle.day_trip_info` is still `None` — otherwise users wouldn't see the sensor until the first successful fetch.
- `state` returns `len(trip_list)` when populated, else `0`.
- `state_attributes` returns `{}` when `day_trip_info` is `None`; otherwise the structure above.
- Times are emitted as naive ISO 8601 (`YYYY-MM-DDTHH:MM:SS`) — friendlier than the raw `YYYYMMDD`/`HHMMSS` strings the library returns.
- Icon: `mdi:calendar`.

`translations/en.json` — adds the `day_trip_info` translation entry.

## Engine-type gate (subtle bug to be aware of)

`vehicle.engine_type` is an `ENGINE_TYPES` enum member, not a string. Comparing it against the strings `"EV"` / `"PHEV"` silently filters every vehicle out. This PR imports `ENGINE_TYPES` from `hyundai_kia_connect_api.const` and compares against `ENGINE_TYPES.EV` / `ENGINE_TYPES.PHEV`.

## Testing

Verified on a real Hyundai Kona Electric (EU):

- Integration reloads cleanly; no `UpdateFailed` in HA logs.
- `sensor.<car>_day_trip_info` registered in the HA entity registry; state goes from `0` → positive integer as trips happen during the day.
- `start_time` / `end_time` render correctly in HA cards (the ISO-8601 conversion was necessary — the raw library strings were unfriendly).
- Tested both the empty-state path (car parked, `NoDataFound` from the endpoint → DEBUG line, sensor stays at `0`) and the populated path (after drives, attributes carry the full trip list and day summary).

## Notes for the maintainer

- `manifest.json` is bumped to a temporary fork-version (`3.2.0-trip.6`) used while iterating on a HACS custom repo. Happy to drop or rename it to whatever clean version you'd like to release this under.
- No new dependencies; all changes use existing library APIs.

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -18,6 +18,7 @@ from hyundai_kia_connect_api import (
     ScheduleChargingClimateRequestOptions,
     Token,
 )
+from hyundai_kia_connect_api.const import ENGINE_TYPES
 from hyundai_kia_connect_api.exceptions import (
     AuthenticationError,
     UnsupportedControlError,
@@ -184,7 +185,10 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
         # a DEBUG log line rather than UpdateFailed.
         today_str = dt_util.now().strftime("%Y%m%d")
         for vehicle_id, vehicle in self.vehicle_manager.vehicles.items():
-            if vehicle.engine_type not in ("EV", "PHEV"):
+            # vehicle.engine_type holds an ENGINE_TYPES enum member, not a
+            # bare string — comparing against ("EV", "PHEV") would never
+            # match (typo silently filtered every vehicle out).
+            if vehicle.engine_type not in (ENGINE_TYPES.EV, ENGINE_TYPES.PHEV):
                 continue
             try:
                 await self.hass.async_add_executor_job(
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -177,6 +177,28 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
                 self.vehicle_manager.update_all_vehicles_with_cached_state
             )
 
+        # Per-trip data (TRIP-01). The library implements update_day_trip_info()
+        # but the upstream coordinator never calls it. Trip data must not fail
+        # the coordinator update — wrap in try/except so a NoDataFound (5921),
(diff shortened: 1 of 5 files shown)


janfrederik:
Superseded by a cleaned-up branch — rebased on current master (3.3.0), dropped the fork-only manifest.json bump, and removed the engine_type filter (the try/except wrapper already handles unsupported endpoints, and the filter risked silently excluding non-EV vehicles where the endpoint actually works). Reopening as a fresh PR from `feat/day-trip-info`.

PR #1698, opened 2026-06-03
feat: add DayTripInfoEntity for per-trip sensor data

janfrederik:
## Summary

Adds a new `DayTripInfoEntity` sensor exposing per-trip data for today, plus the coordinator wiring to populate it. The library (`hyundai_kia_connect_api`) already implements `update_day_trip_info()`, but the upstream coordinator never calls it, so the data is unreachable from HA. This PR closes that gap.

## What it adds

A single new sensor per vehicle (translation key `day_trip_info`, e.g. `sensor.<car>_day_trip_info`):

- **State** — integer trip count for today, `0` if no trips yet, or `unavailable` if the endpoint is not supported for this vehicle/region.
- **Attributes**:
  - `date` — ISO `YYYY-MM-DD` for the day represented.
  - `summary` — day totals: `drive_time`, `idle_time`, `distance`, `avg_speed`, `max_speed` (or `null` before the first successful fetch).
  - `trip_list` — per-trip dicts with `start_time` / `end_time` (naive ISO 8601 local), `drive_time`, `idle_time`, `distance`, `avg_speed`, `max_speed`.

## How it works

`coordinator.py` — in the cached-state branch of `_async_update_data()`, after the existing `update_all_vehicles_with_cached_state()` call:

- Compute `today_str = dt_util.now().strftime("%Y%m%d")` once.
- Loop over `self.vehicle_manager.vehicles`, skipping any already marked unsupported (see below).
- Call `vehicle_manager.update_day_trip_info(vehicle_id, today_str)` for each via `async_add_executor_job`.
- **No exception, `day_trip_info` still `None`** — the library's base-class no-op ran; endpoint not implemented for this region. Vehicle is added to `coordinator.day_trip_unsupported` and skipped on all future polls (resets on HA restart).
- **Exception raised** (e.g. `NoDataFound`) — endpoint exists but returned no data (car parked, no trips yet). Logged at DEBUG; polling continues normally on the next cycle.

`sensor.py` — adds `DayTripInfoEntity(SensorEntity, HyundaiKiaConnectEntity)`:

- Registered unconditionally in `async_setup_entry` (one per vehicle).
- `state` returns `None` (unavailable) when the vehicle is in `coordinator.day_trip_unsupported`, `0` when `day_trip_info` is `None` but the endpoint is supported, or `len(trip_list)` when data is present.
- `state_attributes` returns `{}` when `day_trip_info` is `None`; otherwise the structure above.
- Times are emitted as naive ISO 8601 (`YYYY-MM-DDTHH:MM:SS`) — friendlier than the raw `YYYYMMDD`/`HHMMSS` strings the library returns.
- Icon: `mdi:calendar`.

`strings.json` / `translations/en.json` — adds the `day_trip_info` translation entry.

## Testing

Verified on a real Hyundai Kona Electric (EU):

- Integration reloads cleanly; no `UpdateFailed` in HA logs.
- `sensor.<car>_day_trip_info` registered in the HA entity registry; state goes from `0` → positive integer as trips happen during the day.
- `start_time` / `end_time` render correctly in HA cards (the ISO-8601 conversion was necessary — the raw library strings were unfriendly).
- Tested both the empty-state path (car parked, `NoDataFound` from the endpoint → DEBUG line, sensor stays at `0`) and the populated path (after drives, attributes carry the full trip list and day summary).

No new dependencies; all changes use existing library APIs.

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -178,6 +178,26 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
                 self.vehicle_manager.update_all_vehicles_with_cached_state
             )
 
+        # Per-trip data. The library implements update_day_trip_info() but the
+        # upstream coordinator never calls it. Trip data must not fail the
+        # coordinator update — wrap in try/except so a NoDataFound, an
+        # unsupported region/firmware, or a transient backend error becomes a
+        # DEBUG log line rather than UpdateFailed.
+        today_str = dt_util.now().strftime("%Y%m%d")
+        for vehicle_id in self.vehicle_manager.vehicles:
+            try:
+                await self.hass.async_add_executor_job(
+                    self.vehicle_manager.update_day_trip_info,
+                    vehicle_id,
+                    today_str,
+                )
+            except Exception as exc:
+                _LOGGER.debug(
+                    "Day trip info not available for vehicle %s: %s",
+                    vehicle_id,
+                    exc,
+                )
+
         return self.data
 
     async def async_update_all(self) -> None:
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -65,6 +65,11 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
         """Initialize."""
         self.platforms: set[str] = set()
         self._action_lock = asyncio.Lock()
+        # Vehicle IDs where update_day_trip_info() returned without exception
+        # but left day_trip_info as None — indicates the base-class no-op ran,
+        # meaning the endpoint is not implemented for this region. Resets on
+        # HA restart so a firmware/API change is picked up on next boot.
+        self.day_trip_unsupported: set[str] = set()
 
         self.vehicle_manager = VehicleManager(
             region=config_entry.data.get(CONF_REGION),
@@ -178,22 +183,30 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
                 self.vehicle_manager.update_all_vehicles_with_cached_state
             )
 
-        # Per-trip data. The library implements update_day_trip_info() but the
-        # upstream coordinator never calls it. Trip data must not fail the
-        # coordinator update — wrap in try/except so a NoDataFound, an
-        # unsupported region/firmware, or a transient backend error becomes a
-        # DEBUG log line rather than UpdateFailed.
+        # Per-trip data for today. Skipped for vehicles where the endpoint is
+        # confirmed unsupported (see day_trip_unsupported). On exception the
+        # endpoint exists but returned no data (e.g. no trips yet) — keep
+        # polling. No exception but day_trip_info still None means the library's
+        # base-class no-op ran — endpoint unsupported, stop polling.
         today_str = dt_util.now().strftime("%Y%m%d")
         for vehicle_id in self.vehicle_manager.vehicles:
+            if vehicle_id in self.day_trip_unsupported:
+                continue
             try:
                 await self.hass.async_add_executor_job(
                     self.vehicle_manager.update_day_trip_info,
                     vehicle_id,
                     today_str,
                 )
+                if self.vehicle_manager.vehicles[vehicle_id].day_trip_info is None:
+                    self.day_trip_unsupported.add(vehicle_id)
+                    _LOGGER.debug(
+                        "Day trip info not supported for vehicle %s — skipping future polls",
+                        vehicle_id,
+                    )
             except Exception as exc:
                 _LOGGER.debug(
-                    "Day trip info not available for vehicle %s: %s",
+                    "Day trip info fetch failed for vehicle %s: %s",
                     vehicle_id,
                     exc,
                 )
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -428,9 +428,6 @@ async def async_setup_entry(
         entities.append(
             VehicleEntity(coordinator, coordinator.vehicle_manager.vehicles[vehicle_id])
         )
-        # day_trip_info starts None and is populated by the coordinator's
-        # first update — register unconditionally so the entity exists even
-        # when no trip data has been fetched yet (TRIP-01).
         entities.append(
             DayTripInfoEntity(
                 coordinator, coordinator.vehicle_manager.vehicles[vehicle_id]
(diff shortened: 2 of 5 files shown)


custom_components/kia_uvo/coordinator.py
```diff
@@ -178,6 +178,26 @@ async def _async_update_data(self):
                 self.vehicle_manager.update_all_vehicles_with_cached_state
             )
 
+        # Per-trip data. The library implements update_day_trip_info() but the
+        # upstream coordinator never calls it. Trip data must not fail the
+        # coordinator update — wrap in try/except so a NoDataFound, an
+        # unsupported region/firmware, or a transient backend error becomes a
```
cdnninja: Based on this I wonder if we should have setting to turn this on and off?  As well do we want to be polling this every time using API calls?

custom_components/kia_uvo/sensor.py
```diff
@@ -428,6 +428,14 @@ async def async_setup_entry(
         entities.append(
             VehicleEntity(coordinator, coordinator.vehicle_manager.vehicles[vehicle_id])
         )
+        # day_trip_info starts None and is populated by the coordinator's
+        # first update — register unconditionally so the entity exists even
```
cdnninja: Doesn't this mean it will show up for people with cars that don't support it?

cdnninja (commented):
A few initial thoughts.  Thank you for doing this!

saue0:
I have a Gas Kona 2024, and this endpoint work event if not EV (tested with postman)
https://mybluelink.ca/tods/api/alerts/maintenance/evTripDetails
will this change add the fonctionnality for us as well ?

cdnninja:
> I have a Gas Kona 2024, and this endpoint work event if not EV (tested with postman) [mybluelink.ca/tods/api/alerts/maintenance/evTripDetails](https://mybluelink.ca/tods/api/alerts/maintenance/evTripDetails) will this change add the fonctionnality for us as well ?

It should as trip details are implemented here: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/master/hyundai_kia_connect_api/KiaUvoApiCA.py 

It normalizes to the data model for the regions that implement it.   Are you able to install this patch and test it out?

Merwases:
No, its already available a long time at HA, see Screens:
Screenshot_20260613_200949_Home Assistant
Screenshot_20260613_201025_Home Assistant
Screenshot_20260613_201049_Home Assistant

janfrederik:
Hi @Merwases! Thanks for the screenshots! What you're showing is `sensor.<car>_daily_driving_stats`, which has been in the integration for a while and shows **daily energy totals** (kWh consumed, regenerated, etc.) aggregated by date. That entity calls `update_daily_driving_stats()` under the hood.

This PR adds a different entity: `sensor.<car>_day_trip_info`. It calls the separate `update_day_trip_info()` library function, which the coordinator has never wired up. The new sensor exposes **per-trip detail within a day**: individual trips with `start_time`, `end_time`, distance, average speed, max speed. Not daily energy rollups. The state is the number of trips driven today, not a day count.

So the two live side by side and cover different use cases:

| | `daily_driving_stats` (existing) | `day_trip_info` (this PR) |
|---|---|---|
| State | Number of days of data | Number of trips today |
| Granularity | One row per calendar day | One entry per individual trip |
| Key data | Energy kWh breakdown by date | Per-trip route/timing detail |
| Library call | `update_daily_driving_stats()` | `update_day_trip_info()` |

custom_components/kia_uvo/sensor.py
janfrederik: Good point — addressed in 616d2df. Added `_attr_entity_registry_enabled_default = False` to `DayTripInfoEntity`. Users on unsupported regions/firmware never see a sensor stuck at 0; anyone who wants trip tracking enables it from the HA entity registry once they confirm the endpoint works for their vehicle. Same pattern used elsewhere in HA for optional/region-specific sensors.

custom_components/kia_uvo/coordinator.py
janfrederik: For polling frequency: trip data changes throughout the day as drives happen, so polling on every cached-state cycle is the right cadence — same as `update_daily_driving_stats()`.

For the on/off toggle: addressed in 6691558. The coordinator now detects unsupported vehicles automatically and skips them on future polls — no config option needed. See the other thread for the full explanation.

custom_components/kia_uvo/sensor.py
janfrederik: Reconsidered after discussion — reverting to enabled-by-default (6b7f1c9). The reasoning: disabled-by-default would hide the sensor from the majority of users who can use it, and the discovery mechanism (entity registry) is too obscure for most people to find it. For vehicles where the endpoint is unsupported, the sensor stays at 0 with empty attributes — never goes unavailable, never throws an error. That's the same graceful-degradation pattern `DailyDrivingStatsEntity` already uses, so it sets a consistent expectation. I've added a note to the PR description documenting this behaviour explicitly.

custom_components/kia_uvo/coordinator.py
janfrederik: Updated again in 6691558 with a cleaner approach. The coordinator now distinguishes two failure modes by watching what the library actually does:

- **No exception, `day_trip_info` still `None`** — the base-class `pass` ran; endpoint not implemented for this region. Vehicle is added to `coordinator.day_trip_unsupported` and skipped on all future polls. Sensor returns `unavailable`.
- **Exception raised** (e.g. `NoDataFound`) — a real API call was made but returned no data (car parked, no trips yet today). Logged at DEBUG; polling continues. Sensor returns `0`.

`day_trip_unsupported` is in-memory only, so it resets on HA restart — a firmware or API update that enables the endpoint is picked up on the next boot.

This avoids both the useless-API-call concern and the misleading `0` state for cars that will never have data.

custom_components/kia_uvo/sensor.py
janfrederik: Also updated for this thread in 6691558 — sensor now returns `unavailable` (state `None`) when the vehicle is in `coordinator.day_trip_unsupported`, and `0` only when the endpoint is known to work but has no trips yet today. So users on unsupported cars see an honest `unavailable` rather than a permanent `0`.

taranasus:
Independent confirmation that this works on a **non-EV, EU-region** car — adding it since the PR description only covers a Kona Electric and @saue0's question above (gas Kona, CA) was never closed out with a real result.

**Setup:** Kia Stinger (petrol, no EV hardware at all), Kia Europe account (`brand=1`, `region=1`), HA 2026.8.2, `hyundai_kia_connect_api==4.26.5`. I applied this PR's coordinator + sensor changes by hand on top of integration 3.10.1 (it doesn't apply cleanly to current `main` any more — see the note at the bottom).

**Result:** `update_day_trip_info()` returns full data. Live sensor right now:

```
state: 4
date: 2026-08-21
summary: {'drive_time': 147, 'idle_time': 11, 'distance': 166, 'avg_speed': 50.8, 'max_speed': 157}
trip_list[0]: {'start_time': '2026-08-21T06:02:10', 'end_time': '2026-08-21T06:40:10',
               'drive_time': 37, 'idle_time': 1, 'distance': 61, 'avg_speed': 106, 'max_speed': 157}
```

So `tripinfo` is not EV-gated on the EU endpoint — it's the ICE equivalent of the daily-driving-stats entity that `Merwases` was pointing at, and the distinction @janfrederik drew in the thread holds up: `daily_driving_stats` stays empty on this car while `day_trip_info` populates.

Two observations from running it:

1. **Times are the car's own**, not poll times, which is the real value here. This vehicle polls every 30 min, so trip boundaries derived from odometer deltas were ±30 min; the endpoint gives exact start/end. Worth calling out in the docs — it's the reason to have the entity at all.
2. **`distance` is integer km** in the payload (a 0.6 km trip reports `1`, and one zero-distance record shows up with `avg_speed: 5`), so the day summary won't exactly equal odometer delta. Not a bug in this PR, just worth not documenting as precise.

**One thing blocking merge that isn't code:** the branch is currently in a conflicting state against `main`, and the `coordinator.py` cached-state branch has moved since June. Happy to rebase and push if that helps, @janfrederik / @cdnninja — just say so and I'll open it against the current head.

fellek:
For EU vehicles on the new CCI flow (region "Europe (CCI)"), per-trip data is available from the Kia/Hyundai Driving Insight API, proposed for the library in
Hyundai-Kia-Connect/hyundai_kia_connect_api#1319. It maps well onto this entity:

- `GET /drive-history/trip?vin&date=YYYYMMDD` → `summary` (distance, drive time, daily score)   and `content[]` with `ignOnTime`, `ignOffTime`, `drvDist`, `drvTime`, `idleTime`,  `avgSpeed`, `maxSpeed` — the `TripInfo` fields.
- Each trip also carries `evEnergyUsage` (total, motor, climate, electronics, battery management, regeneration in Wh), which `TripInfo` has no fields for yet.

Two things worth knowing for the entity: the trip timestamps from that API are shifted (one long trip was booked to the next day with its end time as start), and the energy values are 16-bit and can overflow on long trips.

PR #1708, opened 2026-06-08
feat: Add binary sensors for high and bifunc headlamp status

Triple-S:
Hello,

this is a follow up to the addition of the high and bifunc headlamp status to the underlying API repository https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1164. It exposes the newly added status information as new binary sensors like it is done for the preexisting ones.

Best regards,
Triple-S


custom_components/kia_uvo/binary_sensor.py
```diff
@@ -350,6 +350,38 @@ class HyundaiKiaBinarySensorEntityDescription(BinarySensorEntityDescription):
         off_icon="mdi:lightbulb-outline",
         entity_category=EntityCategory.DIAGNOSTIC,
     ),
+    HyundaiKiaBinarySensorEntityDescription(
+        key="headlamp_left_high",
+        translation_key="headlamp_left_high",
```
cdnninja, line 355: Please add the translations to string.json and en.json.

cdnninja (changes requested):
Thanks for doing this!

custom_components/kia_uvo/binary_sensor.py
Triple-S: @cdnninja There are some problems with inconsistent naming conventions in strings.json which is why I hesitated to touch them but since you mentioned it we can try to sort out the problems:

- Are the names supposed to end with "Fault" or not? The names of `headlamp_left_low` and `headlamp_right_low` currently both end with "Fault", however the names of `turn_signal_left_front`, `turn_signal_left_rear`, `turn_signal_right_front` and `turn_signal_right_rear` which are basically the same do not end with "Fault".
- What is more important - Precise naming following a convention or backward compatibility? The names of `headlamp_left_low` and `headlamp_right_low` are currently "Front Left Lamp Fault" and "Front Right Lamp Fault" so the `low` is not reflected in them. If I introduce the `high` and `bifunc` variants is it preferred to keep the existing ones as they are or rename them to reflect the `low`?
- My personal opinion is anyhow that the sensor names should reflect the key names they represent e.g. the sensor for the key `headlamp_left_low` should be named "Headlamp Left Low". This would also ensure that Homeassiatant generates an entity id reflecting the key aka. `binary_sensor.<vehicle_name>_headlamp_left_low` however that would be a quite drastic change compared to the current situation.

Any changes are somewhat critical regarding backward compatibility since Homeassistant generates the entity ids based on the actual names of the sensor (I think this is extremely stupid but it's just the way it works). This will not break anything immediately but once the integration is removed and readded it may result in changed entity ids which is the main reason I hesitate to take a decision in this regard.

cdnninja: I am good either way.    Could just follow your new convention in this pr for new items for now and a follow up to correct others if needed.

Triple-S: I looked through the history to figure out when the first break with entity ids occurred and it was when the translation system was fully intorduced in #1622.

Up to this point the name of `headlamp_left_low` was "Headlamp Left Low Fault" which is reasonably close to the intuitive name based on the key. After this contribution it changed to the current "Front Left Lamp Fault" but unfortunately neigther the PR nor the commit message states the reason for this decision.

@blka As the original author of this change could you explain your thought process behind this change?

I really don't want to introduce further potential breaks in the entity id naming unless there is a good reason to do so or it is inevitable.

blka: Good catch, that was my mistake in #1622. When I migrated from hardcoded `name` to `translation_key`, I accidentally renamed "Headlamp Left Low Fault" → "Front Left Lamp Fault" in strings.json. The original `name` value should have been preserved as-is.

For your new headlamp entities, I'd suggest naming them consistently with the key pattern:
- `headlamp_left_high` → "Headlamp Left High Fault"
- `headlamp_right_high` → "Headlamp Right High Fault"  
- `headlamp_left_bifunc` → "Headlamp Left Bifunc Fault"
- `headlamp_right_bifunc` → "Headlamp Right Bifunc Fault"

And then a follow-up PR can fix the existing `headlamp_left_low` / `headlamp_right_low` names back to "Headlamp Left Low Fault" / "Headlamp Right Low Fault" to match their keys and restore consistency.

The "Fault" suffix makes sense since these binary sensors indicate a fault condition (the lamp is malfunctioning), not that the lamp is simply on.

Triple-S: Thanks for your response. I just committed the changes. I decided to spell out "Bifunctional" because I think this makes more sense than the abbreviation. All translation files besides Polish used the english translation for the existing sensors so I added the new ones there as well.

PR #1721, opened 2026-06-15
feat: add last_scanned_at sensor and debug polling config log

blka:
## Summary

Adds a `last_scanned_at` sensor showing when HA **actually polled the server**, and a debug log at startup showing the effective polling configuration.

### Problem

Users report `last_updated_at` updating every 30 min despite setting `scan_interval = 60 min` (kia_uvo#1718, kia_uvo#1553). The root cause is that `last_updated_at` uses `syncDate` from the API — a server-provided timestamp of when the **car last reported data**, not when HA made the request.

There is currently no way to distinguish:
1. HA is polling too often (misconfiguration)
2. The car reports new data between HA polls (normal telematics)

### Changes

**New sensor: `last_scanned_at`**
- Shows when HA actually polled the server (from `Vehicle.last_scanned_at`, set by the API library)
- `device_class: TIMESTAMP`, `entity_category: DIAGNOSTIC`, icon `mdi:cloud-search`
- Placed right after `last_updated_at` in sensor.py
- Translations added for all 13 languages

**Debug log: polling config at startup**
- Logs `scan_interval`, `force_refresh_interval`, `update_interval`, and `no_force_refresh_hours` when the coordinator initializes
- Makes it trivial to verify settings are applied correctly without digging through options

**Debug log: `_async_update_data` called**
- Logs each `_async_update_data` call with the current intervals
- Helps diagnose double-poll or overlapping cycle issues

### Depends on

- Hyundai-Kia-Connect/hyundai_kia_connect_api#1176 (adds `Vehicle.last_scanned_at` property)
- Manifest version must be bumped after that PR is merged and released

### Related

- kia_uvo#1718
- kia_uvo#1553

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -113,6 +113,17 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
                 seconds=min(self.scan_interval, self.force_refresh_interval)
             ),
         )
+        _LOGGER.info(
+            "%s - Polling configured: scan_interval=%ds, "
+            "force_refresh_interval=%ds, update_interval=%ds, "
+            "no_force_refresh_hours=%d-%d",
+            DOMAIN,
+            self.scan_interval,
+            self.force_refresh_interval,
+            min(self.scan_interval, self.force_refresh_interval),
+            self.no_force_refresh_hour_start,
+            self.no_force_refresh_hour_finish,
+        )
 
     async def _async_update_data(self):
         """Update data via library. Called by update_coordinator periodically.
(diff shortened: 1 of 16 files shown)


custom_components/kia_uvo/coordinator.py
```diff
@@ -113,13 +113,30 @@ def __init__(self, hass: HomeAssistant, config_entry: ConfigEntry) -> None:
                 seconds=min(self.scan_interval, self.force_refresh_interval)
             ),
         )
+        _LOGGER.info(
```
cdnninja: Do we need to log this on every startup? Seems like a debug item?

cdnninja (commented):
One comment on the .info log level.

Issue #1718, opened 2026-06-15
Scan interval not respected and results in a force refresh


LarveyOfficial:
### Region

USA

### Brand of Car

Kia

### Version of the Integration

3.3.2

### Describe the Bug

Similar to #1553 I am seeing a 30 minute interval for the "last updated at" entity.

It seems to respect the no force refresh hours, but after that I am seeing consistent 30 minute checks fire.

I have scan interval set to 60 minutes, and force refresh set to 24 hours. Yet I am still seeing the 30 minute checks fire resulting in my 12V battery getting drained.

### To Reproduce

1. Set scan interval to 60 minutes
2. Set force update interval to 24 hours
3. See last updated at update every 30 minutes

### Expected Behavior

Last updated at should not be updating unless a force update has occurred or the car actually had an update from the server.

I'd also suggest adding a "Last scanned at" entity that indicates the last time the integration actually scanned the kia servers so we can see if this matches the scan interval variable.

### Screenshots

<img width="264" height="60" alt="Image" src="https://github.com/user-attachments/assets/c7977e89-cdfb-4ef9-9c15-37a444f3a369" />

### Additional Context

Seems to be a duplicate of an older closed issue #1553

cdnninja:
Please provide debug logs.

cdnninja:
I wonder if the integration doesn't take the values until you reload it.  I'll have to check to see.  After adjusting the interval could you reload manually?

LarveyOfficial:
Seems that the settings do require a reload in order to actually use them. Tested this by enabling/disabling the encode location with openstreetmap API and noticed that it wouldn't do that unless I reloaded the entire integration.

I will enable debug logging and report back once I get some logs.

Even with the 60 minute interval I am still seeing some irregular battery drain from the car, so I worry that the "scan" is forcing/triggering a force refresh from KIA in some way.

LarveyOfficial:
@cdnninja ~~I am not seeing any debug logs in the system settings. Did something change recently in the way we get debug logs? All I see is 2 errors adding entities and a warning about SSL but no debug logs~~

Edit: never mind i found the logs. Uploading in next comment

LarveyOfficial:
```
2026-06-15 14:20:01.637 DEBUG (SyncWorker_4) [hyundai_kia_connect_api.VehicleManager] hyundai_kia_connect_api - Time differential in seconds: 5246.637132
[...]

PR #1722, opened 2026-06-15
fix: reload on options change

cdnninja:
As per options flow docs using the reload class to trigger a reload.  While we could use listeners I don't think worth the effort. 

Assists with #1718 

@blka  are you able to review for another set of eyes?

### DIFF
--- a/custom_components/kia_uvo/config_flow.py
+++ b/custom_components/kia_uvo/config_flow.py
@@ -150,7 +150,7 @@ async def validate_input(
         raise InvalidAuth from err
 
 
-class HyundaiKiaConnectOptionFlowHandler(config_entries.OptionsFlow):
+class HyundaiKiaConnectOptionFlowHandler(config_entries.OptionsFlowWithReload):
     """Handle an option flow for Hyundai / Kia Connect."""
 
     async def async_step_init(


blka:
LGTM — this fixes a real UX issue. Users changing scan_interval or force_refresh_interval currently have no indication that a reload is needed, and the integration silently keeps using the old values.

A few observations:

**1. This resolves the "30-min interval despite 60-min setting" from #1718**

LarveyOfficial confirmed that after a manual reload, the 60-min interval took effect. `OptionsFlowWithReload` makes this automatic — no more silent mismatch between configured and effective intervals.

**2. The battery drain is a separate issue (Kia server-side, ccNC vehicles)**

From #1050 and the discussion in #1718: on ccNC-head-unit vehicles (2024+ EV9, 2025+ Carnival, 2023 EV6 EU, and likely K5 GT 2026), even `cmm/gvi` (cached state) wakes the car. This is a Kia server bug — the native Kia Access app causes the same wakeups. This PR does not fix that, and nothing on our side can fix it. Users on ccNC vehicles who experience drain should increase `scan_interval` or disable polling entirely.

**3. Commit bba5c51 (`/evc/gts`) is NOT the cause of #1718**

That endpoint is gated on `engine_type in (EV, PHEV)`, and LarveyOfficial's K5 GT is ICE (`fuelType=1`). The `/evc/gts` call is never made for his vehicle.

**4. No debounce, but low risk**

Each "Save" in the options flow triggers a full reload (login + get_vehicles + gvi). If a user changes multiple options quickly, that's multiple reloads in succession. In practice this is unlikely — users don't reconfigure options every minute — but worth noting. If it ever becomes a problem, a short delay before reload would solve it.

**5. Complementary PR: `last_scanned_at` sensor (#1721)**

I have a separate PR open that adds a `last_scanned_at` sensor (when HA actually polled) alongside the existing `last_updated_at` (car's syncDate). This lets users distinguish "HA is polling too often" from "the car reports new data between polls" — which was the core confusion in #1718 even after the reload fix.

Reviewed-by: blka

PR #1745, opened 2026-06-26
feat(svm): SVM image entity + capture button/service for supported vehicles

blka:
## Summary

Expose SVM / Find My Car for Hyundai BlueLink USA in Home Assistant.

Requires `hyundai_kia_connect_api` release containing https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1203.

## What's new

- New `image` entity `SVM Image` showing the latest composite SVM image with metadata attributes.
- New `button` entity `Capture SVM Image` to trigger a fresh capture.
- New service `kia_uvo.capture_svm_image` for automations.
  - Requires `acknowledged_warning=True` (no default).
- Region/capability gated: only USA Hyundai vehicles where `VehicleManager.supports_svm(vehicle_id)` is `True`.

## Privacy

GPS coordinates are not exposed in image attributes; use the existing `device_tracker` entity.

## Notes

- Image entity is created only when capability detection succeeds; it is unavailable until the first image is fetched.
- The `Capture SVM Image` button triggers a capture directly (a button press is an intentional human action). The `kia_uvo.capture_svm_image` service requires `acknowledged_warning=True` to prevent automations from triggering captures in a loop. The asymmetry is deliberate.

### DIFF
--- a/custom_components/kia_uvo/button.py
+++ b/custom_components/kia_uvo/button.py
@@ -13,7 +13,7 @@ from homeassistant.core import HomeAssistant
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
 from hyundai_kia_connect_api import Vehicle
 
-from .const import DOMAIN
+from .const import BRAND_HYUNDAI, DOMAIN, REGION_USA
 from .coordinator import HyundaiKiaConnectDataUpdateCoordinator
 from .entity import HyundaiKiaConnectEntity
 
@@ -92,6 +92,13 @@ BUTTON_DESCRIPTIONS: Final[tuple[HyundaiKiaButtonDescription, ...]] = (
             and vehicle.front_left_window_is_open is not None
         ),
     ),
+    HyundaiKiaButtonDescription(
+        key="capture_svm_image",
+        translation_key="capture_svm_image",
+        icon="mdi:camera-iris",
+        press_action="async_request_svm_capture",
+        exists_fn=lambda _: True,
+    ),
 )
 
 
@@ -105,10 +112,17 @@ async def async_setup_entry(
     for vehicle_id in coordinator.vehicle_manager.vehicles:
         vehicle: Vehicle = coordinator.vehicle_manager.vehicles[vehicle_id]
         for description in BUTTON_DESCRIPTIONS:
-            if description.exists_fn(vehicle):
-                entities.append(
-                    HyundaiKiaConnectButton(coordinator, description, vehicle)
-                )
+            if not description.exists_fn(vehicle):
+                continue
+            if description.key == "capture_svm_image":
+                if (
+                    coordinator.vehicle_manager.region != REGION_USA
+                    or coordinator.vehicle_manager.brand != BRAND_HYUNDAI
+                ):
+                    continue
+                if not await coordinator.async_supports_svm(vehicle_id):
+                    continue
+            entities.append(HyundaiKiaConnectButton(coordinator, description, vehicle))
 
     async_add_entities(entities)
 
--- a/custom_components/kia_uvo/button.py
+++ b/custom_components/kia_uvo/button.py
@@ -13,7 +13,7 @@ from homeassistant.core import HomeAssistant
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
 from hyundai_kia_connect_api import Vehicle
 
-from .const import BRAND_HYUNDAI, DOMAIN, REGION_USA
+from .const import DOMAIN
 from .coordinator import HyundaiKiaConnectDataUpdateCoordinator
 from .entity import HyundaiKiaConnectEntity
 
@@ -97,7 +97,7 @@ BUTTON_DESCRIPTIONS: Final[tuple[HyundaiKiaButtonDescription, ...]] = (
         translation_key="capture_svm_image",
         icon="mdi:camera-iris",
         press_action="async_request_svm_capture",
-        exists_fn=lambda _: True,
+        exists_fn=lambda vehicle: bool(vehicle.supports_svm),
     ),
 )
 
@@ -114,14 +114,6 @@ async def async_setup_entry(
         for description in BUTTON_DESCRIPTIONS:
             if not description.exists_fn(vehicle):
                 continue
-            if description.key == "capture_svm_image":
-                if (
-                    coordinator.vehicle_manager.region != REGION_USA
-                    or coordinator.vehicle_manager.brand != BRAND_HYUNDAI
-                ):
-                    continue
-                if not await coordinator.async_supports_svm(vehicle_id):
-                    continue
             entities.append(HyundaiKiaConnectButton(coordinator, description, vehicle))
 
     async_add_entities(entities)
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -224,10 +224,16 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator[dict[str, Any
         self.async_set_updated_data(self.data)
 
     async def async_supports_svm(self, vehicle_id: str) -> bool:
-        """Return whether the given vehicle supports SVM."""
-        return await self.hass.async_add_executor_job(
-            self.vehicle_manager.supports_svm, vehicle_id
-        )
+        """Return whether the given vehicle supports SVM.
+
+        Capability is a per-region class attribute stamped on the Vehicle by
+        the API library (like supports_window_control), so this is a plain
+        attribute read — no API call, no executor job needed.
+        """
+        vehicle = self.vehicle_manager.vehicles.get(vehicle_id)
+        if vehicle is None:
+            return False
+        return bool(vehicle.supports_svm)
 
     def get_cached_svm_details(self, vehicle_id: str) -> SVMDetails | None:
         """Return cached SVM details for a vehicle, or None if not yet fetched."""
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -67,6 +67,8 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator[dict[str, Any
         self.platforms: set[str] = set()
         self._action_lock = asyncio.Lock()
         self._svm_details: dict[str, SVMDetails] = {}
+        # Per-vehicle SVM fisheye dewarp toggle (local UI state, off by default).
+        self._svm_dewarp: dict[str, bool] = {}
 
         self.vehicle_manager = VehicleManager(
             region=config_entry.data.get(CONF_REGION),
@@ -234,6 +236,14 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator[dict[str, Any
         if vehicle is None:
             return False
         return bool(vehicle.supports_svm)
+
+    def svm_dewarp_enabled(self, vehicle_id: str) -> bool:
+        """Return the per-vehicle SVM fisheye dewarp preference."""
+        return self._svm_dewarp.get(vehicle_id, False)
+
+    def set_svm_dewarp(self, vehicle_id: str, enabled: bool) -> None:
+        """Set the per-vehicle SVM fisheye dewarp preference."""
+        self._svm_dewarp[vehicle_id] = enabled
 
     def get_cached_svm_details(self, vehicle_id: str) -> SVMDetails | None:
         """Return cached SVM details for a vehicle, or None if not yet fetched."""
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -229,6 +229,10 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator[dict[str, Any
             self.vehicle_manager.supports_svm, vehicle_id
         )
 
+    def get_cached_svm_details(self, vehicle_id: str) -> SVMDetails | None:
+        """Return cached SVM details for a vehicle, or None if not yet fetched."""
+        return self._svm_details.get(vehicle_id)
+
     async def async_get_svm_details(self, vehicle_id: str) -> SVMDetails:
         """Fetch the latest cached SVM image and metadata from the API."""
         details = await self.hass.async_add_executor_job(
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -252,7 +252,7 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator[dict[str, Any
         details = await self.hass.async_add_executor_job(
             self.vehicle_manager.request_svm_capture,
             vehicle_id,
-            True,  # acknowledged_warning — enforced by the capture service
+            True,  # acknowledged_warning — capture is always a user-initiated action
         )
         self._svm_details[vehicle_id] = details
         self.async_set_updated_data(self.data)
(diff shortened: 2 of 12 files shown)


custom_components/kia_uvo/coordinator.py
```diff
@@ -65,6 +66,7 @@ def __init__(self, hass: HomeAssistant, config_entry: ConfigEntry) -> None:
         """Initialize."""
         self.platforms: set[str] = set()
         self._action_lock = asyncio.Lock()
+        self._svm_details: dict[str, SVMDetails] = {}
```
cdnninja, line 69: Why would we store here vs on vehicle?

custom_components/kia_uvo/coordinator.py
```diff
@@ -221,6 +223,36 @@ async def async_force_refresh_vehicle(self, vehicle_id: str) -> None:
...
+        """Return whether the given vehicle supports SVM."""
+        return await self.hass.async_add_executor_job(
+            self.vehicle_manager.supports_svm, vehicle_id
+        )
+
+    async def async_get_svm_details(self, vehicle_id: str) -> SVMDetails:
+        """Fetch the latest cached SVM image and metadata from the API."""
+        vehicle = self.vehicle_manager.vehicles[vehicle_id]
+        details = await self.hass.async_add_executor_job(
+            self.vehicle_manager.api.get_svm_details,
```
cdnninja: So far I don't think we call the API layer directly.

custom_components/kia_uvo/coordinator.py
```diff
@@ -221,6 +223,36 @@ async def async_force_refresh_vehicle(self, vehicle_id: str) -> None:
...
+        return details
+
+    async def async_request_svm_capture(self, vehicle_id: str) -> SVMDetails:
+        """Trigger a fresh SVM capture and update the cached details."""
+        vehicle = self.vehicle_manager.vehicles[vehicle_id]
+        details = await self.hass.async_add_executor_job(
+            self.vehicle_manager.api.request_svm_capture,
+            self.vehicle_manager.token,
+            vehicle,
+            True,  # acknowledged_warning
```
cdnninja: If we plan to just force true is it worth even having is as a value?

custom_components/kia_uvo/coordinator.py
blka: Because the image payload doesn't belong on `Vehicle` — see my reply on the API PR. Short version: ~170 KB JPG + GPS-bearing metadata, on-demand not polled. `Vehicle` carries the `supports_svm` capability flag; the image bytes are cached by HA's `ImageEntity` (and held on the coordinator between captures). Keeping the JPG off `Vehicle` avoids bloating every state dump/snapshot.

custom_components/kia_uvo/coordinator.py
blka: Agreed — this should go through `VehicleManager`, not `manager.api` directly. I'm adding `VehicleManager.get_svm_details` / `request_svm_capture` wrappers on the API PR (#1203); once that lands I'll switch the coordinator to call them and this `manager.api` reach-through goes away.

custom_components/kia_uvo/coordinator.py
blka: The parameter still has value at the API layer — library consumers calling `request_svm_capture` directly must acknowledge. In HA the safety gate is one level up: the `kia_uvo.capture_svm_image` service requires `acknowledged_warning: true` with no default, so by the time the coordinator runs the user has already acknowledged. Hardcoding `True` here just avoids re-checking what the service already enforced. I don't want to drop the parameter from the API — it's the library's safety contract.

custom_components/kia_uvo/coordinator.py
blka: Done — both PRs updated.

- API #1203: added `VehicleManager.get_svm_details(vehicle_id)` and `VehicleManager.request_svm_capture(vehicle_id, acknowledged_warning)` wrappers (commit 4d14d57).
- This PR: coordinator now calls `self.vehicle_manager.get_svm_details(...)` / `request_svm_capture(...)`; the `manager.api` reach-through and manual token/vehicle passing are gone (commit 7dfa816).

As before, this PR depends on the API library release containing #1203

custom_components/kia_uvo/button.py
```diff
@@ -105,10 +112,17 @@ async def async_setup_entry(
...
         for description in BUTTON_DESCRIPTIONS:
-            if description.exists_fn(vehicle):
-                entities.append(
-                    HyundaiKiaConnectButton(coordinator, description, vehicle)
-                )
+            if not description.exists_fn(vehicle):
+                continue
+            if description.key == "capture_svm_image":
+                if (
+                    coordinator.vehicle_manager.region != REGION_USA
```
cdnninja: Do we need both or can we focus on supports SVM?

custom_components/kia_uvo/image.py
```diff
@@ -0,0 +1,88 @@
...
+    config_entry: ConfigEntry,
+    async_add_entities: AddEntitiesCallback,
+) -> None:
+    """Set up image platform."""
+    coordinator: HyundaiKiaConnectDataUpdateCoordinator = hass.data[DOMAIN][
+        config_entry.unique_id
+    ]
+
+    if (
+        coordinator.vehicle_manager.region != REGION_USA
```
cdnninja: Same as the other, can we just trust the API responds correctly? Otherwise this needs changing as the API adds support.

custom_components/kia_uvo/coordinator.py
```diff
@@ -221,6 +223,41 @@ async def async_force_refresh_vehicle(self, vehicle_id: str) -> None:
...
+        Capability is a per-region class attribute stamped on the Vehicle by
+        the API library (like supports_window_control), so this is a plain
+        attribute read — no API call, no executor job needed.
+        """
+        vehicle = self.vehicle_manager.vehicles.get(vehicle_id)
+        if vehicle is None:
+            return False
+        return bool(vehicle.supports_svm)
+
+    def get_cached_svm_details(self, vehicle_id: str) -> SVMDetails | None:
```
cdnninja, line 248: Should this be async?

custom_components/kia_uvo/button.py
```diff
@@ -105,10 +112,17 @@ async def async_setup_entry(
...
+                    or coordinator.vehicle_manager.brand != BRAND_HYUNDAI
+                ):
+                    continue
+                if not await coordinator.async_supports_svm(vehicle_id):
```
cdnninja: Can't this be the lambda?

custom_components/kia_uvo/services.py
```diff
@@ -331,6 +333,21 @@ async def async_handle_set_navigation(call: ServiceCall) -> None:
         )
         await coordinator.async_set_navigation(vehicle_id, [poi])
 
+    async def async_handle_capture_svm_image(call: ServiceCall) -> None:
+        coordinator = _get_coordinator_from_device(hass, call)
+        vehicle_id = _get_vehicle_id_from_device(hass, call)
+        acknowledged_warning = call.data.get("acknowledged_warning")
```
cdnninja: So we bypass this with the button but prompt on the service?

cdnninja (changes requested):

custom_components/kia_uvo/button.py
blka: Right — dropped the region/brand special-case. The button now uses `exists_fn` reading `vehicle.supports_svm`, which the API library stamps from the per-region class attribute; new regions come free with library updates.

custom_components/kia_uvo/image.py
blka: Agreed — the gate is gone. Setup now checks `vehicle.supports_svm` (stamped by the API library), so support for new regions arrives with library updates and no integration change.

custom_components/kia_uvo/coordinator.py
blka: Intentionally sync — it's a plain dict read with no I/O, and it's called from the sync `available` property and `_handle_coordinator_update` on the event loop, which can't await. The async API calls live in `async_get_svm_details` / `async_request_svm_capture`.

custom_components/kia_uvo/button.py
blka: Done — moved to `exists_fn=lambda vehicle: bool(vehicle.supports_svm)`; the special-case block in the setup loop is gone (cfbd690).

custom_components/kia_uvo/services.py
blka: Fair point — gate removed (cfbd690). Both the button press and the service call are explicit user-initiated actions; the `acknowledged_warning` safety contract stays at the API layer, and the field is gone from the service schema.

custom_components/kia_uvo/svm_dewarp.py
```diff
@@ -0,0 +1,147 @@
+"""Fisheye dewarp for SVM camera views.
```
cdnninja, line 1: Would this be a good utility to add to the library so other consumers get the benefit?

badnbusy:
@blka  I have a UK (EU) Kona SX2 (2024 onwards) that supports SVM. Can I use/test this? It works in the MyHyundai app...

ijojog:
SVM capture is working on my Hyundai USA 2025 Tucson Hybrid, but the SVM image entities fail to render in Home Assistant 3.13.0.

The vehicle receives the request and the MyHyundai app shows the new capture, but all HA SVM image entities display broken images.

Home Assistant logs show the cause consistently:

AttributeError: 'SVMDetails' object has no attribute 'image_sizes'. Did you mean: 'image_size'?

Traceback points to:

custom_components/kia_uvo/image.py line 124

Current code is checking:

details.image_sizes

but the SVMDetails object returned by the API appears to expose:

details.image_size

So the capture/cache succeeds, but async_image() fails before the image can be served.

This repeats for every SVM image request.

I can test a fix on my live Hyundai USA vehicle if needed.

blka:
Thanks for the detailed report — this one is on us, a release-chaining mistake.

**Root cause:** the SVM code merged in 3.13.0 reads `SVMDetails.image_sizes`, added to the API lib in 4.31.0 (Hyundai-Kia-Connect/hyundai_kia_connect_api#1310). The 3.13.0 manifest still pins `4.30.0`, which predates that field. That's why capture works (it only needs `image_bytes`) and the render crashes on the missing attribute. The manifest bump to 4.31.0 is already on master (#1883) and ships with the next release.

**Workaround until then:**

1. Edit `/config/custom_components/kia_uvo/manifest.json`:
   ```json
   "requirements": ["hyundai_kia_connect_api==4.31.0"]
   ```
2. Restart HA — it installs the pinned version on startup.

Two caveats: a bare `pip install 4.31.0` won't stick — HA enforces the manifest pin and downgrades back on restart. And the next HACS update restores the 4.30.0 pin, so re-apply after updating.

And yes — a check on your live vehicle after applying the workaround would be welcome. The camera crops index `image_sizes` (cameras use index 2, bird's-eye index 4), which is exactly the code path 4.30.0 can't serve.

ijojog:
Thanks — confirmed working on my live US 2025 Tucson Hybrid after changing the manifest requirement to hyundai_kia_connect_api==4.31.0 and restarting Home Assistant.

SVM capture now works and the bird’s-eye, front, rear, left, and right image entities all render correctly. The camera crops also appear correct.

I also confirmed that the Dewarp switch works. Initially I thought it needed to be set before capture, but it turns out the image update just takes a while after toggling Dewarp. Once it refreshes, the difference is visible.

Happy to continue testing any SVM changes on the live Hyundai USA vehicle.

blka:
Not yet, unfortunately — and thanks for checking. On the EU region the integration currently uses the legacy EU API class, which doesn't expose SVM (`supports_svm` stays false), so the entities and the capture button won't even be created for your vehicle. The EU SVM implementation lives in the library's Europe-CCI implementation, which the integration doesn't expose as a selectable region yet. For now SVM works for Hyundai BlueLink USA vehicles only.

That README row was misleading for EU users — it describes library capability, not what the integration wires up today. I'll get it corrected. When the CCI region lands in the integration, EU vehicles like yours are exactly the target use case.

cdnninja:
I was thinking.  Should dewarp be a config setting rather than a switch?

ijojog:
I like it the easy you've done it. 

I'm working on making an interactive panoramic viewer and will share with you if I can do it.

custom_components/kia_uvo/image.py
```diff
@@ -0,0 +1,176 @@
...
+        self._attr_image_last_updated = details.captured_at if details else None
+        super()._handle_coordinator_update()
+
+    @property
+    def available(self) -> bool:
+        """Return True if a cached image is available."""
+        return self.coordinator.get_cached_svm_details(self.vehicle.id) is not None
+
+    async def async_image(self) -> bytes | None:
+        """Return bytes of this view cropped from the latest SVM composite."""
```
fuatakgun, line 122: Nit: i would personally move this cropping/splitting logic into library so consumers can access these images directly or else, everyone needs to do the same work over and over.

fuatakgun (commented):
Dropped a nit

fuatakgun:
Thanks for doing this

ijojog:
Since this PR has already been merged and closed, I just wanted to share one additional idea in case you ever feel it would be useful to incorporate into the SVM feature later.

I am not an engineer, and I am definitely not the ideal person to do camera calibration or image-projection work. I only experimented with this because I wanted a more useful way to view the individual SVM camera images in Home Assistant, and since I managed to get something working for myself, I thought the idea might also be useful to other users.

What I ended up doing was creating two separate custom cards:

a 3D surround-view experiment, using the four SVM camera images plus the bird’s-eye image;
and a much simpler SVM camera gallery/viewer, which I actually think may be more practical.

The gallery displays the Front, Rear, Left, Right and Bird’s-eye SVM images individually. When you click one of the camera images, it opens in a larger viewer where you can zoom in and move around the image.

For the four fisheye camera images, I also experimented with applying a fisheye-to-perspective / rectilinear transformation. The idea is that instead of always looking at the heavily curved “bubble” image, when you zoom into an area and move around it, the selected part of the image is reprojected into a much more normal-looking perspective.

I implemented this as separate JavaScript Lovelace resources because that was simply the only way I knew how to experiment with it. I am sure you would not need to use my resources or my exact implementation at all. You could probably implement the same concept much more cleanly directly inside the integration/card if you ever wanted to.

So I am mainly sharing the idea, rather than suggesting that my code should be adopted as-is:

show all SVM cameras individually;
click a camera to open it larger;
allow zoom and pan;
optionally dewarp the fisheye image into a perspective view;
keep the raw image available as a toggle;
include the Bird’s-eye view;
and potentially offer a surround/3D view separately for users who want it.

The 3D stitching/calibration side is obviously much more difficult, and my version is still experimental. But the individual-camera viewer with perspective correction has been very useful for me, especially when trying to inspect details that are difficult to see in the original fisheye image.

Since the hard work of exposing the SVM images is already in the main version now, I thought this might be a useful future enhancement for people using those entities.

If useful, I can also post the example YAML and the JavaScript I used for the experiments. Again, they are only proof-of-concept code, and someone who understands image calibration/projection properly could certainly make them much better.  Images are concepts so you understand the idea, but not how they actually show on my end.

SVM CAMERAS


PANORAMIC VIEW


How I implemented the proof of concept

For my own Home Assistant installation, I created two small custom Lovelace cards:

tucson-svm-gallery.js — displays the individual Front, Right, Rear, Left and Bird's-eye SVM images. Clicking a camera opens a larger interactive view where the fisheye image can be explored, zoomed and panned.
tucson-svm-panorama.js — experimental panoramic viewer that uses the four SVM camera images to create an interactive surround view. The user can rotate around the vehicle, zoom into areas of interest, and switch between an orbit-style view and a view closer to the edge/camera position.

I added these as JavaScript modules under /config/www/ and registered them as Home Assistant dashboard resources under /local/.

I am not suggesting that this exact implementation or calibration should be incorporated into the integration. I am not an engineer and this was mostly experimentation using the images returned by my own Tucson. I am sure the developers here could implement the concept much more cleanly and accurately.

My main reason for sharing it is that once the integration exposes the individual SVM camera images, there is a lot more that can potentially be done with them than simply displaying the five images individually.

Even just an official/simple gallery with click → fullscreen → pan/zoom would make the SVM functionality substantially more useful in Home Assistant. The panoramic viewer is an additional experimental idea for what could eventually be possible.

If useful, I can also share the JavaScript and Lovelace YAML I used for the proof of concept.

PR #1755, opened 2026-06-28
fix: TypeError when in-car temperature is a non-numeric value

aaronwolen:
I noticed the climate entity periodically crashes the coordinator listener update with:

```
TypeError: Temperature is not a number: 72
  File ".../homeassistant/components/climate/__init__.py", line 355, in state_attributes
    ATTR_CURRENT_TEMPERATURE: show_temp(...)
  File ".../homeassistant/helpers/temperature.py", line 23, in display_temp
    raise TypeError(f"Temperature is not a number: {temperature}")
```

Home Assistant's climate base class passes `current_temperature`/`target_temperature` through `display_temp()`, which throws for any non-numeric value. I ran into this because the US backends (`KiaUvoApiUSA`, `HyundaiBlueLinkApiUSA`) set `vehicle.air_temperature` without casting, unlike the EU/other backends, which use `float(air_temp)`). So for US vehicles the string flows into `display_temp()` and crashes the listener update on every coordinator refresh.

## Fix

This PR coerces the temperature values to `float` in the climate entity and returns `None` for missing or non-numeric values. It also corrects the `hvac_mode` / `hvac_action` comparisons, which previously compared values lexically as strings (e.g. `"72" > "70"`).

## Testing

I tested on my local Home Assistant instance with a US Kia Telluride and confirmed `climate.*_climate_control` reports `current_temperature` / `temperature` as numbers and no `TypeError` appears in the logs.

### DIFF
--- a/custom_components/kia_uvo/climate.py
+++ b/custom_components/kia_uvo/climate.py
@@ -110,16 +110,31 @@ class HyundaiKiaCarClimateControlSwitch(HyundaiKiaConnectEntity, ClimateEntity):
             return UnitOfTemperature(self.vehicle._air_temperature_unit)
         return UnitOfTemperature.CELSIUS
 
+    @staticmethod
+    def _as_float(value) -> float | None:
+        """Coerce a temperature value to float.
+
+        Some API backends (e.g. the US ones) report the in-car temperature as a
+        string such as "72" instead of a number. Home Assistant's climate base
+        class rejects non-numeric temperatures with a TypeError, so coerce here.
+        """
+        if value is None:
+            return None
+        try:
+            return float(value)
+        except (TypeError, ValueError):
+            return None
+
     @property
     def current_temperature(self) -> float | None:
         """Get the current in-car temperature."""
-        return self.vehicle.air_temperature
+        return self._as_float(self.vehicle.air_temperature)
 
     @property
     def target_temperature(self) -> float | None:
         """Get the desired in-car target temperature."""
         # TODO: use Coordinator data, not internal state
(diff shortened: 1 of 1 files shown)


cdnninja:
It feels like we should fix this library side?

blka:
Heads up: the root cause of this `TypeError` lives in the upstream API library, not in `kia_uvo`.

`Vehicle.air_temperature` (and `outside_temperature`) is set without a `float()` cast by the US backends (`KiaUvoApiUSA`, `HyundaiBlueLinkApiUSA`) and BR, while `ApiImplType1` (EU/CCS2) does cast. So for US vehicles the raw string (`"72"`) flowed into the `Vehicle` object and onward to HA's `display_temp()`, which rejects non-numeric values.

API-side fix: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1209 — coerces both fields to `float | None` at the `Vehicle` setter level (single choke point every backend passes through). Includes a US regression test and updates the US snapshots.

Once #1209 merges and `kia_uvo` bumps its pinned `hyundai_kia_connect_api` version, the `_as_float` wrapper added here becomes redundant — the value is already `float | None`. It stays safe in the meantime (the `if value is None: return None` guard means `float(None)` is never reached), so this PR can either land as-is for defense-in-depth or be closed in favor of the API fix. Either is fine; just flagging the dependency so reviewers have the full picture.

aaronwolen:
Thanks to you both. I haven't had a chance to look at the api library, so I appreciate you taking up the torch, @blka. I'll close this PR.

PR #1760, opened 2026-06-29
fix(sensor): use EV battery instead of 12V on device info page

### LINKED ISSUE #1749: Wrong battery level being shown on device info page (shows 12v rather than EV battery)
### Region

USA

### Brand of Car

Hyundai

3.5.0

### Describe the Bug

On the device info page, it is showing the 12v battery as the battery in the header instead of the EV battery level. The EV battery is more correct to use here.

Image

_No response_

### To Reproduce

1. Have an EV
2. Go to device page, compare the battery shown at top to EV battery level sensor

### Expected Behavior

For electric vehicles, it should use the EV Battery level (sensor.2026_ioniq_5_ev_battery_level) rather than the 12v battery level (sensor.2026_ioniq_5_car_battery_level)

### Screenshots

_No response_

### Additional Context

_No response_

blka:
Fixes #1749.

Marks `car_battery_percentage` sensor with `EntityCategory.DIAGNOSTIC` so Home Assistant no longer picks the 12 V battery as the device's battery indicator. For EVs, the device page will now show the EV battery level sensor instead.

### DIFF
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -72,6 +72,7 @@ SENSOR_DESCRIPTIONS: Final[tuple[SensorEntityDescription, ...]] = (
         native_unit_of_measurement=PERCENTAGE,
         device_class=SensorDeviceClass.BATTERY,
         state_class=SensorStateClass.MEASUREMENT,
+        entity_category=EntityCategory.DIAGNOSTIC,
     ),
     SensorEntityDescription(
         key="last_updated_at",


FunkyTob:
On my side, I still see the Car battery percentage. The only thing that changed is the icon.

blka:
Thanks for reporting back. The fix landed in v3.5.1 — `car_battery_percentage` is now marked `EntityCategory.DIAGNOSTIC`, while `ev_battery_percentage` stays as the primary battery entity. Home Assistant should then pick the EV battery for the device header.

A couple of things to check, since the icon changing but the header staying on 12 V points at the integration not being fully reloaded rather than a code issue:

1. **Full HA restart** after updating to v3.5.1? Reloading the integration via UI is not enough — `entity_category` is a static property baked into the entity registry at platform startup, and HA often restores the old attributes from cache until a full restart.
2. Does the `ev_battery_percentage` sensor exist and have a real state (not `unknown`/`unavailable`)? If it has no value, HA falls back to the only battery entity with a state — which would be the 12 V one.
3. HA version?

If you can share a screenshot from **Developer Tools → States** for both `sensor.*_car_battery_percentage` and `sensor.*_ev_battery_percentage`, that will tell us whether it's an install/cache issue or something we need to fix in the integration.

Issue #1790, opened 2026-07-14
set_temperature sensor unavailable when airTemp value is OFF/LO — entity not created at startup


pipedexplumbing:
### Region
USA

### Brand of Car
Hyundai

### Version of the Integration
v3.7.1 (hyundai_kia_connect_api v4.24.1)

### Describe the Bug

The `set_temperature` sensor (entity key `_air_temperature`) becomes permanently `unavailable` (`restored: True`) after a Home Assistant restart when the vehicle climate is off.

**Root cause:** In `sensor.py` line 455:
```python
if getattr(vehicle, description.key, None) is not None:
    entities.append(HyundaiKiaConnectSensor(coordinator, description, vehicle))
```

The entity is only created if `vehicle._air_temperature` is not `None` at startup. When climate is off, Hyundai USA API returns:
```json
"airTemp": {"unit": 1, "hvacTempType": 1, "value": "OFF"}
```

In `HyundaiBlueLinkApiUSA.py` (line 462-469):
```python
air_temp = get_child_value(state, "vehicleStatus.airTemp.value")
if air_temp == "LO":
    air_temp = self.temperature_range[0]   # 62
if air_temp == "HI":
    air_temp = self.temperature_range[-1]  # 82
if air_temp:
    vehicle.air_temperature = (air_temp, TEMPERATURE_UNITS[1])
```

**The `"OFF"` case is not handled.** When `air_temp == "OFF"`:
- It does not match `"LO"` or `"HI"`
- `if air_temp:` is truthy (string "OFF" is truthy)
- `float_or_none("OFF")` returns `None` via the Vehicle.py setter
- `vehicle._air_temperature` becomes `None`
- The sensor entity is NOT created at startup
- Entity remains `restored: True` / `unavailable` permanently

Even after climate is turned on and API returns `"LO"` (maps to 62), the entity is never re-evaluated for creation because `async_setup_entry` only runs once.

### To Reproduce
1. Ensure vehicle climate is OFF
2. Restart Home Assistant
3. `sensor.<vehicle>_set_temperature` shows `unavailable` with `restored: True`
4. Turn climate on via Bluelink app — sensor remains `unavailable`
5. Force refresh does not help — entity was never created

### Expected Behavior
Sensor should be created even when climate is off, showing `62` (LO mapping) or a numeric value, not `unavailable`/`restored`.

### Workaround
1. Turn climate ON via Bluelink
[...]

PR #1791, opened 2026-07-15
fix: always create air_temperature sensor entity (#1790)

blka:
## Summary

Special-case `_air_temperature` in the sensor entity-creation gate so the `set_temperature` sensor is always created, instead of only when `vehicle._air_temperature` is not `None` at startup.

## Why

The `_air_temperature` sensor was only created when `vehicle._air_temperature` was not `None` at startup (`sensor.py` `async_setup_entry`):

```python
for description in SENSOR_DESCRIPTIONS:
    if getattr(vehicle, description.key, None) is not None:
        entities.append(HyundaiKiaConnectSensor(coordinator, description, vehicle))
```

When climate is off, the USA backend returns `airTemp.value == "OFF"` and `air_temperature` is `None`, so the entity was **never created**. Since `async_setup_entry` runs once at startup, the entity stayed permanently `unavailable` (`restored: True`) — even after climate was turned on, which is why the reporter's only recovery was turning climate on + `reload_config_entry` + force refresh.

## What changed

```python
for description in SENSOR_DESCRIPTIONS:
    # _air_temperature is None while climate is off (USA returns
    # airTemp.value "OFF"); create the entity anyway so it reports
    # `unknown` instead of being permanently unavailable after a
    # restart. The real setpoint arrives on the next poll.
    if (
        description.key == "_air_temperature"
        or getattr(vehicle, description.key, None) is not None
    ):
        entities.append(HyundaiKiaConnectSensor(coordinator, description, vehicle))
```

The `_air_temperature` entity is created for every vehicle. When `_air_temperature is None` (climate off), `native_value` returns `None` → HA state `unknown` (the HA-correct representation of "setpoint not known"). When climate turns on, the coordinator pushes the real setpoint on the next poll — no reload needed. All other sensors keep the `is not None` gate (EV/fuel/etc. still skip when unsupported).

Climate is a core vehicle feature; all Hyundai/Kia Bluelink vehicles have climate, so no spurious entity is created. The `climate_control` entity is unaffected — it gates on `air_control_is_on` (a boolean present even when off), so it was already always created.

## Not done (deliberate)

- **No masking:** the API returns `"OFF"` (it does not preserve the last setpoint in this field), so the setpoint is genuinely unknown while climate is off. `unknown` is Home Assistant's honest representation; reporting 62°F (mapping `OFF`→`LO`) would be a fake setpoint. When climate is turned back on, the real value appears on the next update.
- **No `manifest.json` change** — this is a kia_uvo-internal change with no API dependency change. The companion API PR (skip-`OFF` conformance) releases separately and the bump pipeline picks it up.

## Tests

kia_uvo has no standalone test suite (verified via HA Docker). The change is a one-line gate special-case. Verification: restart HA with vehicle climate off → `sensor.<vehicle>_air_temperature` shows `unknown` (not `unavailable`/`restored: True`); turn climate on via Bluelink → next coordinator update shows the numeric setpoint without `reload_config_entry`. ruff clean.

Refs: Hyundai-Kia-Connect/kia_uvo#1790

### DIFF
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -452,14 +452,20 @@ async def async_setup_entry(
     for vehicle_id in coordinator.vehicle_manager.vehicles.keys():
         vehicle: Vehicle = coordinator.vehicle_manager.vehicles[vehicle_id]
         for description in SENSOR_DESCRIPTIONS:
-            # _air_temperature is None while climate is off (USA returns
-            # airTemp.value "OFF"); create the entity anyway so it reports
-            # `unknown` instead of being permanently unavailable after a
-            # restart. The real setpoint arrives on the next poll.
-            if (
-                description.key == "_air_temperature"
-                or getattr(vehicle, description.key, None) is not None
-            ):
+            if description.key == "_air_temperature":
+                # The setpoint is transient — it is None while climate is off
+                # (USA returns airTemp.value "OFF"), so don't gate on it. Gate
+                # on climate presence (air_control_is_on, the same signal the
+                # climate entity uses) to avoid creating an unusable sensor on
+                # vehicles that report no climate. A None setpoint -> HA
+                # `unknown`; the real setpoint arrives on the next poll.
+                create = (
+                    vehicle.air_control_is_on is not None
+                    or vehicle._air_temperature is not None
+                )
+            else:
+                create = getattr(vehicle, description.key, None) is not None
+            if create:
                 entities.append(
                     HyundaiKiaConnectSensor(coordinator, description, vehicle)
                 )
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -452,7 +452,14 @@ async def async_setup_entry(
     for vehicle_id in coordinator.vehicle_manager.vehicles.keys():
         vehicle: Vehicle = coordinator.vehicle_manager.vehicles[vehicle_id]
         for description in SENSOR_DESCRIPTIONS:
-            if getattr(vehicle, description.key, None) is not None:
+            # _air_temperature is None while climate is off (USA returns
+            # airTemp.value "OFF"); create the entity anyway so it reports
+            # `unknown` instead of being permanently unavailable after a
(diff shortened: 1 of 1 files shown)


cdnninja:
Are we confident all cars report this data or would this create a unusable sensor for some of our users?

blka:
Good point — thanks. I checked the data rather than assuming:

- All 10 region handlers (USA Hyundai/Kia, EU, EU CCS2, CA, CN, AU, IN, BR, CCI) set `air_control_is_on`.
- Every test fixture across regions reports `air_control_is_on` as a non-None value (`False`/`0` while climate is off) — including the EU EV9 on the CCS2 path (`Cabin.HVAC.*`), which still yields `air_control_is_on=0`. So a Bluelink vehicle reports the climate on/off state even when climate is off.

That said, "always create" was too blunt. I've refined the gate (30b8a68): instead of creating the `_air_temperature` entity unconditionally, it's created when the vehicle **reports climate** (`air_control_is_on is not None` — the same signal the `climate_control` entity already gates on in `climate.py`) **or already has a setpoint**:

```python
if description.key == "_air_temperature":
    create = (
        vehicle.air_control_is_on is not None
        or vehicle._air_temperature is not None
    )
else:
    create = getattr(vehicle, description.key, None) is not None
if create:
    entities.append(HyundaiKiaConnectSensor(coordinator, description, vehicle))
```

A vehicle that reports neither climate state nor a setpoint (no climate) gets no sensor — so no unusable entity for climate-less vehicles. This is a strict superset of the original `air_temperature is not None` gate, so there's no regression for any vehicle that currently shows a setpoint.

The original bug is still fixed: a climate-off vehicle has `air_control_is_on=False` (not `None`), so the sensor is created at startup, reports `unknown` while the setpoint is unknown, and shows the real value on the next poll — without `reload_config_entry`.

Verified against fixtures + simulated cases:

| case | air_control_is_on | air_temperature | original gate | refined gate |
|---|---|---|---|---|
| climate on + setpoint (fixtures) | False/0 | set | created | created (no regression) |
| climate off (#1790) | False | None | **not created (bug)** | **created → `unknown`** |
| no climate | None | None | not created | not created (no unusable sensor) |

Issue #1778, opened 2026-07-09
Australia: Authentication fails


domwolfy:
### Region

AUS

### Brand of Car

Hyundai

### Version of the Integration

3.6.0

### Describe the Bug

I had the integration working until yesterday, then it said it needed to be re-configured which failed. 

Tried restart of home assistant, removing the integration completely (confirming nothing in the custom_components folder) and after re-install it's the same error. Have also tried logging out and in of the official app. 

on the UI the error is "Login failed into Hyundai.....servers...Please use the official app to log out and log back in..." 

system logs below. 

### Debug Logs (if an error occurred)

2026-07-07 01:33:49.664 INFO (MainThread) [homeassistant.setup] Setting up kia_uvo
2026-07-07 01:33:49.664 INFO (MainThread) [homeassistant.setup] Setup of domain kia_uvo took 0.00 seconds
2026-07-07 01:33:50.067 ERROR (SyncWorker_7) [hyundai_kia_connect_api.ApiImplType1] Unknown API response format: {'errId': '95c34ec3-2b50-48b0-b6f1-252d2f503b50', 'errCode': '4002', 'errMsg': 'Invalid parameters'}
2026-07-07 01:33:50.067 WARNING (SyncWorker_7) [hyundai_kia_connect_api.ApiImplType1] hyundai_kia_connect_api - Refresh token exchange failed, falling back to full login
2026-07-07 01:33:51.514 WARNING (MainThread) [homeassistant.config_entries] Config entry 'Kia Australia kia25@dojawo.slmail.me' for kia_uvo integration could not authenticate: Login Failed
2026-07-07 01:59:23.915 ERROR (SyncWorker_5) [hyundai_kia_connect_api.ApiImplType1] Unknown API response format: {'errId': 'd8d036a8-b5eb-4a61-982f-e9710155c9b5', 'errCode': '4002', 'errMsg': 'Invalid parameters'}
2026-07-07 01:59:23.916 WARNING (SyncWorker_5) [hyundai_kia_connect_api.ApiImplType1] hyundai_kia_connect_api - Refresh token exchange failed, falling back to full login
2026-07-07 01:59:24.894 WARNING (MainThread) [custom_components.kia_uvo] Config entry 'Kia Australia kia25@dojawo.slmail.me' for kia_uvo integration could not authenticate: Login Failed
[...]

Issue #1786, opened 2026-07-13
Force refresh error


morton027:
### Region

EU

### Brand of Car

Hyundai

### Version of the Integration

3.7.0

### Describe the Bug

Force reload button has stopped working. Attempting to reload Bluelink data returns the following error:

Failed to perform the action button/press. 'resMsg'

I attempted this due to the tyre pressure readings of 255 psi. Note that I haven’t driven the car since updating, which may be a cause of the incorrect pressures.

### Debug Logs (if an error occurred)

Logger: homeassistant.components.websocket_api.http.connection
Source: components/websocket_api/commands.py:281
Integration: Home Assistant WebSocket API (documentation, issues)
First occurred: 08:57:30 (7 occurrences)
Last logged: 09:13:52

[546705290432] Unexpected exception
[546334260128] Unexpected exception
[546364487168] Unexpected exception
[546376513984] Unexpected exception
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 281, in handle_call_service
    response = await hass.services.async_call(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<7 lines>...
    )
    ^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2900, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2943, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 949, in entity_service_call
    response_data = await _async_handle_entity_calls(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
    )
    ^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 828, in _async_handle_entity_calls
    single_result = await entity.async_request_call(_with_context(entity, coro))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1708, in async_request_call
    return await coro
           ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 823, in _with_context
[...]

Issue #1231, opened 2025-08-07
EU Hyundai Login Issue


Maggi0r:
Please check Services, Known Bug / Issues and Troubleshooting over here first: https://github.com/fuatakgun/kia_uvo/blob/master/README.md
**Region and Brand of car**

**2.43.4**

**Config Not Ready: 'token_type'**
At 4 am the Last Connect works at after this Cennection nothing works.

**Screenshots**
![Image](https://github.com/user-attachments/assets/4de9bec3-6617-406a-8526-afd181338b6a)

Regards Maggi

Bitzecoinze:
Same issue at the same time (4am in central Europe)

<img width="305" height="448" alt="Image" src="https://github.com/user-attachments/assets/5f3418f7-36f1-4558-8cba-8d9736152388" />

paulschermers:
The same error here. Hereby my log details if that is helpfull:
```
Logger: custom_components.kia_uvo.coordinator
Source: helpers/update_coordinator.py:380
integration: Hyundai / Kia Connect ([documentation](https://github.com/Hyundai-Kia-Connect/kia_uvo), [issues](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues))
First occurred: 03:55:13 (11 occurrences)
Last logged: 06:34:57

Unexpected error fetching kia_uvo data
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 380, in _async_refresh
    self.data = await self._async_update_data()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/kia_uvo/coordinator.py", line 109, in _async_update_data
    await self.async_check_and_refresh_token()
  File "/config/custom_components/kia_uvo/coordinator.py", line 171, in async_check_and_refresh_token
    await self.hass.async_add_executor_job(
        self.vehicle_manager.check_and_refresh_token
    )
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.13/site-packages/hyundai_kia_connect_api/VehicleManager.py", line 153, in check_and_refresh_token
    self.token: Token = self.api.login(self.username, self.password)
                        ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1796, opened 2026-07-19
fix: create CCS2 tire pressure sensors when the structure exists, not only when a value is present

stuartdenne:
Numeric tire-pressure sensors are gated on the value being non-None at integration setup.  On the AU/NZ backend TPMS is transient: the cached state carries the 255 no-data sentinel (-> None) whenever the car is parked - which is nearly always the case when HA starts - so for these cars the sensors are never created at all, and users only ever see the binary warning-lamp sensors.

Live-confirmed on Kia NZ (region 7, same account as #1778): parked/early-drive snapshots carry `Pressure: 255` on all four wheels; a snapshot a few minutes into a drive carries real values (37 psi x 4); the last driving snapshot then persists in the cached state after ignition-off (verified >=20 min), so the regular 30-min poll picks up real values after essentially every drive - the sensors are genuinely useful once they exist.

Fix mirrors the `_air_temperature` precedent: gate creation on the CCS2 tire structure being present in the raw data (`_tire_pressure_sensor_supported`), OR the value being present (unchanged behaviour for backends that always report values).  A None pressure shows as `unknown` until a poll catches the car driving.

Tests: `tests/test_tire_sensor_creation.py` (structure-present-parked -> create; non-CCS2 -> don't; CCS2-without-structure -> don't).  `ruff check` + `format` clean; no new failures in the committed suite.  Running live on my install now.

Disclosure, as with hyundai_kia_connect_api#1242: largely Claude (AI) work - reviewed, tested and live-run by me.

### DIFF
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -472,6 +472,20 @@ async def async_setup_entry(
                 # next reload. Always create; None -> HA `unknown`, the real
                 # SoC arrives on the next poll. See #1803.
                 create = True
+            elif description.key.startswith("tire_pressure_"):
+                # Transient like _air_temperature above: some backends (AU/NZ)
+                # report the TPMS no-data sentinel whenever the car is parked
+                # — nearly always the case at setup — so don't gate on the
+                # value. The parsed unit is the capability signal: non-None
+                # exactly for direct-TPMS vehicles (known PressureUnit), None
+                # for indirect TPMS (PressureUnit 3, e.g. KONA — #1786) and
+                # old-protocol vehicles, which never report a numeric
+                # pressure. A None pressure -> HA `unknown` until a poll
+                # catches the car driving.
+                create = (
+                    getattr(vehicle, description.key, None) is not None
+                    or vehicle.tire_pressure_unit is not None
+                )
             else:
                 create = getattr(vehicle, description.key, None) is not None
             if create:
--- a/tests/test_tire_sensor_creation.py
+++ b/tests/test_tire_sensor_creation.py
@@ -1,0 +1,100 @@
+"""Tests for numeric tire-pressure sensor creation gating.
+
+On backends where TPMS is transient (confirmed live on AU/NZ), the cached
+state carries the 255 no-data sentinel (parsed to None) whenever the car is
+parked — nearly always the case at integration setup — so gating creation on
+the value would mean the sensors never exist. Creation is gated on the parsed
+``vehicle.tire_pressure_unit`` instead: non-None exactly for direct-TPMS
+vehicles, None for indirect-TPMS vehicles (PressureUnit 3, e.g. the KONA EV
+dump in kia_uvo #1786) and old-protocol vehicles, which can never report a
+numeric pressure.
+
+The payload fixtures run through the REAL API-library CCS2 parser so these
+tests pin the whole contract, not a re-implementation of it.
+"""
+
+from unittest.mock import MagicMock
+
+from hyundai_kia_connect_api import Vehicle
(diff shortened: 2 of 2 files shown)


blka:
True, was about to check this.

blka:
Nice — mirrors the `_air_temperature` always-create precedent from #1791, and `unknown` for a None pressure is the HA-correct representation. CCS2 scope and the helper tests look right.

One concern with the gate: `_tire_pressure_sensor_supported` checks `Chassis.Axle.Row1.Left.Tire.Pressure` is present in `vehicle.data`. That's also true for **indirect-TPMS** vehicles — e.g. the KONA EV dump in kia_uvo #1786 returns `PressureUnit: 3` with `Pressure: 0` on all wheels (no direct sensor; pressure inferred from wheel-speed differences). The API library parses unit 3 to `tire_pressure_unit = None` (#1231) and the values stay `None`, so this gate would create 4 numeric sensors that show `unknown` **forever** on a car that can never report a pressure.

The cleaner "has direct TPMS" signal is the parsed unit: `vehicle.tire_pressure_unit is not None` — non-None exactly for known units (0/1/2) and `None` for unit 3 (indirect) or no TPMS field. It also conforms to the #1791 pattern (which gates on parsed `vehicle._air_temperature`, not raw `vehicle.data`) and drops the helper + `get_child_value` import:

```python
elif description.key.startswith("tire_pressure_"):
    create = (
        getattr(vehicle, description.key, None) is not None
        or vehicle.tire_pressure_unit is not None
    )
```

For AU/NZ parked (your case): `PressureUnit` stays 0/1/2 in the cached state (it's the car's display-unit setting, independent of whether sensors reported), so `tire_pressure_unit` is non-None and sensors are created — same outcome as your helper, without the indirect-TPMS false-positive. A test with `PressureUnit: 3` (KONA shape) asserting no creation would pin this.

custom_components/kia_uvo/sensor.py
```diff
@@ -441,6 +442,25 @@
...
+    sentinel (parsed to None) whenever the car is parked — which is nearly
+    always the case at integration setup, so gating on the value would mean
+    the sensors are never created. Gate on the CCS2 tire structure being
+    present in the raw data instead. Real values arrive with the first
+    poll that lands during or shortly after a drive.
+    """
+    if not vehicle.ccu_ccs2_protocol_support:
+        return False
+    return (
+        get_child_value(vehicle.data, "Chassis.Axle.Row1.Left.Tire.Pressure")
```
cdnninja: This is region specific so should occur in the regions.

cdnninja (changes requested):

stuartdenne:
Good point - didn't realise indirect TPMS was a thing with the Kona.  I've been testing my original PR for the past week and it's already been useful to establish that a recent screw incident hasn't caused a serious leak.  A bit messy to parse as the sensor values are dropped when not driving, but I'm reluctant to try and persist / guess at values given the variability in pressure with temperature and so on.

- Gate is now the parsed `vehicle.tire_pressure_unit is not None`, OR'd with the existing value check.  No raw-payload access, no helper, no `get_child_value` import.
- Verified in the library source that the unit parses to None for both PressureUnit 3 (indirect TPMS) and absent/old-protocol.  The region implementations also show CN/IN never invoke the CCS2 parser at all, so the gate stays conservative there too.
- Tests now run the real library CCS2 parser over two payload shapes - the AU/NZ parked shape (PressureUnit 0, 255 sentinel -> 4 sensors created) and the #1786 KONA shape (PressureUnit 3 -> none) - then assert actual creation through `async_setup_entry`, so they pin the contract rather than a re-implementation.  Third case: a vehicle never parsed through the CCS2 path -> none.
- Rebased onto master (picks up the #1804 12V-battery always-create case landing in the same setup loop).

PR #1808, opened 2026-07-24
feat: time entities for off-peak charging window (#1764)

blka:
## Depends on `hyundai_kia_connect_api` #1251 — ✅ merged & released (4.26.0)

#1251 was merged and released as `hyundai_kia_connect_api==4.26.0`; kia_uvo `master` is already pinned to 4.26.0 (#1826), so this PR is now unblocked and the `ev_off_peak_*` attributes it reads are populated by the library.

## Summary

Part of #1764 — exposes and controls the off-peak charging schedule from Home Assistant.

- **`time.py`** — new `TimeEntity` platform with `ev_off_peak_start_time` and `ev_off_peak_end_time`, following the existing description pattern. `exists_fn` gates on `is not None`, consistent with `switch.py`/`sensor.py`.
- **`coordinator.py`** — `async_set_off_peak_charging(*, mode=None, start, end)` building on `_build_schedule_options_from_vehicle` + `async_schedule_charging_and_climate`. `mode=None` preserves the current mode (time-entity path); the mode maps to the `(charging_enabled, off_peak_charge_only_enabled)` pair the API expects.
- **`set_off_peak_charging` service** — convenience preset: a single `mode` selector (Off / Time priority / Target priority) + optional start/end window, with a `description` (hassfest allows `description` on services, unlike on entities) that carries the time-vs-target priority explanation. Non-breaking additive service; the existing `schedule_charging_and_climate` low-level service and the switch/time entities are unchanged.
- **`strings.json` + `en.json` + `pl.json`** — names for the two time entities. Other language files fall back to `strings.json` (existing sparse convention; `pl.json` time-entity names are English placeholders to be translated). Entity entries use only `name` (hassfest rejects `description` on entities).
- **`__init__.py`** — register `Platform.TIME`.

The library side (`schedule_charging_and_climate`, including the EV5-appMode flat-endpoint dispatch for ccNC EVs) lives in `hyundai_kia_connect_api` #1251 — this PR only consumes it.

## Note on the existing `schedule_charging_and_climate` service

The pre-existing `schedule_charging_and_climate` service is a **full-reset** service: it builds the request from the call data only (not from the current vehicle state), and omitted fields fall back to defaults (e.g. `charging_enabled` defaults to `false`). **Callers must specify every field they want kept** — omitting `charging_enabled` disables scheduled charging, and omitting the departure fields disables the departure presets.

Now that #1251 makes the service actually apply for ccNC/EV5 vehicles (previously the combined `/ccs2/reservation/chargehvac` was a no-op for them), this full-reset behaviour is active for those vehicles too. The new **`set_off_peak_charging` preset service** and the **time entities** added here are the **partial-update** path: they preserve the current schedule via `_build_schedule_options_from_vehicle` and override only the field(s) being changed, so they are safe to call without resetting the rest of the schedule. Prefer them for partial changes; use the low-level service only when you intend to set the whole schedule at once.

## Pre-merge checks (self-audit)

1. **HA principles**: `exists_fn` None-gates, consistent with `switch.py`/`sensor.py`. No wake on a cheap path; writes go through the coordinator.
2. **Pattern conformance**: time entity + coordinator helpers + service mirror existing patterns (`open_charge_port` description, `schedule_charging_and_climate` handler). Translations follow the existing sparse convention.
3. **No bloat**: focused additions; no English duplicated across language files (fallback); the preset service reuses existing coordinator plumbing.
4. **Structural hooks**: reuses `_build_schedule_options_from_vehicle` + `async_schedule_charging_and_climate`; the low-level `schedule_charging_and_climate` service and the switch are kept.
5. **Right layer**: entities in `time.py`, write orchestration in `coordinator.py`, service wiring in `services.py`.
6. **Type safety**: full hints; `mode` validated in the coordinator (`ValueError` on unknown).
7. **Surgical scope**: additive (new platform + service + helpers); no existing service/entity changed — non-breaking.
8. **No dead code**.

## Test plan

- [x] `ruff check` + `ruff format` — clean
- [x] `python -m py_compile` on `time.py`/`coordinator.py`/`services.py`/`const.py`/`__init__.py` — OK
- [x] YAML + JSON validity — OK
- [x] `pre-commit run` (ruff/codespell/pyupgrade/yamllint/prettier) — pass
- [x] `validate-hassfest` — pass (no `description` on entities; service `description` allowed)
- [ ] Functional: pending live test against 4.26.0

### DIFF
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -446,6 +446,22 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
         options.off_peak_charge_only_enabled = enabled
         await self.async_schedule_charging_and_climate(vehicle_id, options)
 
+    async def async_set_off_peak_time(
+        self,
+        vehicle_id: str,
+        *,
+        start: dt.time | None = None,
+        end: dt.time | None = None,
+    ) -> None:
+        """Set the off-peak charging window start and/or end time."""
+        vehicle = self.vehicle_manager.vehicles[vehicle_id]
+        options = self._build_schedule_options_from_vehicle(vehicle)
+        if start is not None:
+            options.off_peak_start_time = start
+        if end is not None:
+            options.off_peak_end_time = end
+        await self.async_schedule_charging_and_climate(vehicle_id, options)
+
     async def async_set_departure_enabled(
         self, vehicle_id: str, departure_num: int, enabled: bool
     ):
--- a/custom_components/kia_uvo/coordinator.py
+++ b/custom_components/kia_uvo/coordinator.py
@@ -462,6 +462,40 @@ class HyundaiKiaConnectDataUpdateCoordinator(DataUpdateCoordinator):
             options.off_peak_end_time = end
         await self.async_schedule_charging_and_climate(vehicle_id, options)
 
+    async def async_set_off_peak_charging_mode(
+        self,
+        vehicle_id: str,
+        mode: str,
+        *,
+        start: dt.time | None = None,
+        end: dt.time | None = None,
+    ) -> None:
+        """Set the off-peak charging schedule mode and optionally the window.
+
+        ``mode`` maps to the (charging_enabled, off_peak_charge_only_enabled)
+        pair the API expects: ``off`` disables scheduled charging, ``time``
+        charges only during the off-peak window (time priority), ``target``
+        prefers off-peak tariffs but continues past the window to reach the
+        target SoC (target priority). Other schedule fields are preserved.
+        """
+        vehicle = self.vehicle_manager.vehicles[vehicle_id]
+        options = self._build_schedule_options_from_vehicle(vehicle)
+        if mode == "off":
+            options.charging_enabled = False
+        elif mode == "time":
+            options.charging_enabled = True
+            options.off_peak_charge_only_enabled = True
+        elif mode == "target":
+            options.charging_enabled = True
+            options.off_peak_charge_only_enabled = False
+        else:
+            raise ValueError(f"Invalid off-peak charging mode: {mode!r}")
+        if start is not None:
+            options.off_peak_start_time = start
+        if end is not None:
+            options.off_peak_end_time = end
+        await self.async_schedule_charging_and_climate(vehicle_id, options)
+
     async def async_set_departure_enabled(
         self, vehicle_id: str, departure_num: int, enabled: bool
     ):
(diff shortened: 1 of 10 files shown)


custom_components/kia_uvo/coordinator.py
```diff
@@ -448,6 +448,56 @@ async def async_set_off_peak_charge_only_enabled(
...
+        if start is not None:
+            options.off_peak_start_time = start
+        if end is not None:
+            options.off_peak_end_time = end
+        await self.async_schedule_charging_and_climate(vehicle_id, options)
+
+    async def async_set_off_peak_charging_mode(
+        self,
+        vehicle_id: str,
+        mode: str,
```
cdnninja: Looks like mode should be an enum?

custom_components/kia_uvo/coordinator.py
```diff
@@ -448,6 +448,56 @@ async def async_set_off_peak_charge_only_enabled(
...
+        """Set the off-peak charging window start and/or end time."""
+        vehicle = self.vehicle_manager.vehicles[vehicle_id]
+        options = self._build_schedule_options_from_vehicle(vehicle)
+        if start is not None:
+            options.off_peak_start_time = start
+        if end is not None:
+            options.off_peak_end_time = end
+        await self.async_schedule_charging_and_climate(vehicle_id, options)
+
+    async def async_set_off_peak_charging_mode(
```
cdnninja: Could these methods be combined?  Seems they are identical except mode.

cdnninja:
Overall looks good! few comments.

custom_components/kia_uvo/coordinator.py
blka: Done in 7e6c7b7 — added `OffPeakChargingMode(StrEnum)` in `const.py` with `OFF="off"`, `TIME="time"`, `TARGET="target"`. The service handler wraps the payload value with `OffPeakChargingMode(mode)`, so an invalid value raises `ValueError` at the boundary instead of falling through silently, and the coordinator now takes the enum directly.

custom_components/kia_uvo/coordinator.py
blka: Done in 7e6c7b7 — merged into one `async_set_off_peak_charging(vehicle_id, *, mode=None, start=None, end=None)`. `mode=None` preserves the current mode and only adjusts the window, which is exactly the path the time entities need (they call it without a mode). The `services.py` handler passes `OffPeakChargingMode(mode)`; `time.py` calls it with only `start=`/`end=`. Drops one coordinator method, no behaviour change.

PR #1821, opened 2026-07-30
fix(climate): unit-aware min/max temp for Fahrenheit vehicles

blka:
## Summary

`climate.py` hardcoded `min_temp=14` / `max_temp=30` — the Celsius range. For vehicles that report Fahrenheit (USA/CA, `air_temperature_unit == "°F"`), the climate entity therefore exposed a **14-30 °F** slider, making `set_temperature` via the entity unusable. Derive the bounds from the existing `temperature_unit` property so Fahrenheit vehicles get 62-82 °F.

Celsius behaviour is unchanged (14-30).

## Change

`custom_components/kia_uvo/climate.py` — `min_temp` / `max_temp` now branch on `self.temperature_unit == UnitOfTemperature.FAHRENHEIT` (62/82) vs Celsius (14/30). Uses the already-populated `vehicle._air_temperature_unit`; no API-library change.

## Scope / caveats

- This is a unit-aware stopgap. The exact per-region range (Kia USA `62-82`, Hyundai USA `62-82`, EU `17-30` half-step, etc.) is still a TODO best sourced from the library; this PR only fixes the broken Fahrenheit case without duplicating the full `temperature_range` table.
- `target_temperature_step` is still hardcoded `0.5` (°C); for °F it should be `1`. Intentionally out of scope here to keep the change surgical — noted for a follow-up.
- No open issue references this exact symptom (the USA °F min/max); it's a latent correctness bug surfaced during a climate-code audit. If a maintainer prefers the fuller lib-sourced range instead, happy to defer to that.

## Testing

kia_uvo has no unit-test suite (HA custom component; HA not importable in CI). Verified:
- `ruff check` + `ruff format` clean.
- `python -m py_compile` clean.
- pre-commit hooks (ruff, codespell, prettier) pass.
- Logic: `temperature_unit` already returns `UnitOfTemperature` from `vehicle._air_temperature_unit`; F → 62/82, C → 14/30 (unchanged).

### DIFF
--- a/custom_components/kia_uvo/climate.py
+++ b/custom_components/kia_uvo/climate.py
@@ -128,18 +128,22 @@ class HyundaiKiaCarClimateControlSwitch(HyundaiKiaConnectEntity, ClimateEntity):
         # TODO: get from lib
         return 0.5
 
-    # TODO: unknown
     @property
     def min_temp(self) -> float:
         """Get the minimum settable temperature."""
-        # TODO: get from lib
+        # TODO: get the exact per-region range from the lib
+        # USA/CA report Fahrenheit; the hardcoded 14-30 °C bounds made the
+        # climate slider unusable (14-30 °F) for those vehicles.
+        if self.temperature_unit == UnitOfTemperature.FAHRENHEIT:
+            return 62
         return 14
 
-    # TODO: unknown
     @property
     def max_temp(self) -> float:
         """Get the maximum settable temperature."""
-        # TODO: get from lib
+        # TODO: get the exact per-region range from the lib
+        if self.temperature_unit == UnitOfTemperature.FAHRENHEIT:
+            return 82
         return 30
 
     @property


blka:
Review update with the temperature-range rationale.

The climate entity `min_temp` / `max_temp` should match the range the **library** accepts and clamps `set_temp` to — not the app's wider UI picker range, because the library is what actually receives and clamps the value from HA.

Library `temperature_range` (the authoritative bounds for `set_temp`):
- Hyundai USA: `range(62, 82)` → **62–82 °F**
- Kia USA: `range(62, 83)` → 62–82 °F
- EU / `ApiImplType1`: `0.5 * range(28, 60)` → **14.0–29.5 °C**

The official app's UI picker exposes a wider Fahrenheit range (≈57–91 °F on the legacy table, 58–88 °F on the newer one), but the library clamps `set_temp` to its own `temperature_range`, so the HA entity bounds must follow the library range to avoid the user picking a value HA accepts but the library then rejects/clamps. Hence:
- Fahrenheit (`temperature_unit == °F`, USA/CA): `62` / `82` ✓ matches the library's USA range
- Celsius: `14` / `30` ✓ matches the library's EU range (14.0–29.5)

This is a unit-aware stopgap. The longer-term fix is to expose the per-region `temperature_range` from the library so the entity reads the exact bounds (including CA's old/new tables and Kia vs Hyundai USA) — kept as a TODO in the code, intentionally out of scope here.

PR #1835, opened 2026-08-04
fix: keep EV charging power sensor available

jimlau:
### Summary

- always create the EV charging-power sensor for EV and PHEV vehicles
- retain value-based discovery for backends that report charging power without
  an engine type
- keep unsupported empty charging-power entities off ICE vehicles
- add regression coverage for EV, PHEV, ICE, and unknown-engine scenarios

### Background

Charging power is transient and is normally `None` when a vehicle is unplugged.
The sensor platform currently creates most entities only when their value is
non-`None` during integration setup. This means the charging-power sensor is not
loaded when Home Assistant starts while the vehicle is unplugged. A later
coordinator poll can contain a valid power reading, but no entity exists to
publish it; reloading the integration while charging makes it appear.

The value was confirmed in an anonymized Hyundai USA payload under
`vehicleStatus.evStatus.batteryStndChrgPower`, while the registered Home
Assistant entity remained unavailable. After creating the sensor based on the
stable EV/PHEV capability, the entity changed to the current kW value at setup
and can receive subsequent coordinator updates normally. When no power is
reported, Home Assistant represents the sensor as unknown rather than removing
it from the loaded platform.

### Testing

- EV with `None` power: sensor is created with an unknown value
- PHEV with `None` power: sensor is created
- ICE with `None` power: sensor is not created
- unknown engine type with reported power: existing discovery behavior remains
- all four scenarios passed against Home Assistant 2026.7.4 with the patched
  integration
- `ruff check`, `ruff format --check`, and `git diff --check` passed

### Privacy

The tests use synthetic values and identifiers. No VIN, account identifier,
coordinates, authentication data, or raw personal vehicle payload is included.

### DIFF
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -470,6 +470,15 @@ async def async_setup_entry(
                 # next reload. Always create; None -> HA `unknown`, the real
                 # SoC arrives on the next poll. See #1803.
                 create = True
+            elif description.key == "ev_charging_power":
+                # Charging power is transient and usually None at setup while
+                # the vehicle is unplugged. Create the sensor for electrified
+                # vehicles so a later coordinator poll can publish the value
+                # without requiring an integration reload.
+                create = (
+                    vehicle.engine_type in (ENGINE_TYPES.EV, ENGINE_TYPES.PHEV)
+                    or vehicle.ev_charging_power is not None
+                )
             elif description.key.startswith("tire_pressure_"):
                 # Transient like _air_temperature above: some backends (AU/NZ)
                 # report the TPMS no-data sentinel whenever the car is parked
--- a/tests/test_ev_charging_power_sensor_creation.py
+++ b/tests/test_ev_charging_power_sensor_creation.py
@@ -1,0 +1,59 @@
+"""Tests for transient EV charging-power sensor creation."""
+
+from unittest.mock import MagicMock
+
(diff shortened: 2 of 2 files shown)


sethfitz (approved):
Independent corroboration — I hit this exact bug on a 2025 Ioniq 5 (Hyundai USA path) and, before finding this PR, had written the same fix on a branch of my own. The creation gate here matches it, and I can confirm the two subtle points that make it correct:

- **Gating on `engine_type in (EV, PHEV)` is the right stable signal.** Charging power is transient (`None` while unplugged), so a value-only gate drops the entity on any reload taken while unplugged — which is how I originally misread it as a release having removed the sensor. Drivetrain is stable across charge state. This mirrors the `#1803` `car_battery_percentage` precedent, but correctly stops short of an unconditional `create = True`: charging power is meaningless on ICE/HEV, so an always-`unknown` sensor there would be its own regression.
- **Keeping `or vehicle.ev_charging_power is not None` as a fallback is load-bearing, not defensive.** `KiaUvoApiCA` can leave `engine_type` unset, so gating on drivetrain alone would create *fewer* entities than today for those users. The `or` keeps the change strictly additive — worth guarding against a later "simplification" that drops it.
- **Leaving `ev_charging_current` untouched is the right scope.** It's assigned only in `ApiImplType1` (the CCS2 path), never by either USA implementation, so gating it the same way would hand every USA user a permanently-`unknown` sensor. Different bug; correct to keep it out.

LGTM.

Issue #1842, opened 2026-08-13
Target Range of Charge AC / DC not available since last update


JWartenberg:
### Region

EU

### Brand of Car

Hyundai

### Version of the Integration

3.10.0

### Describe the Bug

Entities "Target Range of Charge AC" and "Target Rage of Charge DC" are not longer available since last update.

### To Reproduce

Display Sensors: they are marked as not available

### Expected Behavior

Values should be visible.

danielTinkiel:
Hello JWartenberg, I have a Ioniq 6 2026 and my Hyundai integration is also v3.10. The entity sensor.garage_ioniq6_target_range_of_charge_ac which is the "Target Range of Charge AC" is giving me a normal response,  currently 581 KM in my case. Also the entity sensor.garage_ioniq6_target_range_of_charge_dc, which is the "Target Range of Charge DC" is also responding properly, 516KM in my case. I hope this response will help you to narrow down the issue origin, whether it is specific to your model or something else. Kind regards, Daniel

gerosil:
i'm on 3.10.0, too and have for my Kona the same issue. The following values are since 3.10 always the same & refer in my opinion to the current possible distance (kona_ev_range):

kona_target_range_of_charge_ac
kona_target_range_of_charge_dc
kona_ev_range
kona_total_driving_range

updated on 3.10.1, same issue. here are the readings:


drvDistance:
      - rangeByFuel:
          evModeRange:
            value: 270
            unit: 1
          totalAvailableRange:
            value: 270
            unit: 1
        type: 2


targetSOClist:
        - targetSOClevel: 100
          dte:
            rangeByFuel:
              evModeRange:
                value: 270
                unit: 1
              totalAvailableRange:
                value: 270
                unit: 1
            type: 2
          plugType: 0
        - targetSOClevel: 100
          dte:
            rangeByFuel:
              evModeRange:
                value: 270
                unit: 1
              totalAvailableRange:
                value: 270
                unit: 1
            type: 2
          plugType: 1

JWartenberg:
> "Target Range of Charge AC" is giving me a normal response, currently 581 KM in my case. 

Congratulations. I haven't reached a figure like that yet (around 560 at most).
The problem seems to be fixed in the next update.

Thanks to blka.

gerosil:
should it work now in 3.13.0? i have still the same values in kona_target_range_of_charge_ac, kona_target_range_of_charge_dc and kona_ev_range

PR #1847, opened 2026-08-18
fix(sensor): keep EV target-charge-range sensors available (#1842)

blka:
## Summary

The **Target Range of Charge AC** and **Target Range of Charge DC** sensors became unavailable for some EU Hyundai/Kia EVs after the update to 3.10.x (Hyundai-Kia-Connect/kia_uvo#1842).

This gates those two sensors on the stable EV/PHEV capability instead of on their value being non-`None` at setup, mirroring the charging-power sensor (#1835) and the 12V battery sensor (#1803).

## Root cause

Both sensors fall into the generic `else` branch of `async_setup_entry`:

```python
else:
    create = getattr(vehicle, description.key, None) is not None
```

The range is derived from `evStatus.reservChargeInfos.targetSOClist[i].dte.rangeByFuel.totalAvailableRange.value`. That value is **transient** — it is `None` at setup when:

- the telematics unit is asleep / not reporting `dte.rangeByFuel`,
- the car is parked and the backend omits the range block,
- or `targetSOClist` has a single entry, so one of the two hardcoded indices (`.0.` / `.1.`) is absent.

When a version-update reload lands while the value is `None`, the entity is **not yielded**, HA marks it "no longer provided", and it only comes back on a manual reload that happens to coincide with the value being present. This is the same bug class as the 12V battery sensor (#1803) and the EV charging-power sensor (#1835).

The API-library parsing is **unchanged** across `v4.25.6 → v4.26.5` (diffed `KiaUvoApiEU.py` / `Vehicle.py` — no edits to the `targetSOClist` read path). The regression is the HA-side value-gated creation, not parsing.

## Fix

Gate the two target-charge-range sensors on the stable EV/PHEV capability with the existing value-based fallback — identical pattern to the charging-power sensor (#1835) — now expressed as an `exists` predicate on the sensor description (the same pattern the binary_sensor platform uses for `is_on`):

```python
HyundaiKiaSensorEntityDescription(
    key="_ev_target_range_charge_AC",
    ...
    exists=lambda vehicle: (
        _is_electrified(vehicle) or vehicle._ev_target_range_charge_AC is not None
    ),
),
```

`async_setup_entry` consumes it with a single check (`description.exists(vehicle)` when defined, otherwise the default value gate), replacing the per-key elif cascade. A `None` value → HA `unknown` until the next coordinator poll publishes the range. ICE/HEV vehicles are unaffected.

## Scope

- **In:** EV and PHEV vehicles (which report a target charge range).
- **Out:** ICE / HEV (do not report this range).
- Applies uniformly to all regions that parse `targetSOClist` (EU / USA / AU / CN / IN share the same `else` gate); strictly more permissive only for EV/PHEV, no regression for other powertrains.

## Secondary symptom (not fixed here)

One commenter (Kona EV) reports the AC and DC values are equal to the current range. Their diagnostics payload shows both `targetSOClist` entries carry the same `totalAvailableRange` from the backend — i.e. the car's API returns the current range for both plug types. The parser reads the field correctly; this is car-side data, not a parsing bug, and is out of scope for this PR.

## Testing

- `tests/test_ev_target_range_sensor_creation.py` — regression coverage mirroring the charging-power tests:
  - EV with `None` range → both sensors created, value `None` (→ HA `unknown`)
  - PHEV with `None` range → both sensors created
  - ICE with `None` range → neither created
  - unknown engine type with a reported range → sensors created with the reported value
- Full local suite: 66 passed (Home Assistant + pytest-asyncio, Python 3.13).
- `ruff check` and `ruff format --check` clean on all tracked `.py`.
- No change to `manifest.json`, `strings.json`, or `services.yaml` → hassfest / HACS unaffected.

### DIFF
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -3,6 +3,8 @@
 from __future__ import annotations
 
 import logging
+from collections.abc import Callable
+from dataclasses import dataclass
 from datetime import date
 from typing import Final
 
@@ -31,8 +33,21 @@ from .entity import HyundaiKiaConnectEntity
 
 _LOGGER = logging.getLogger(__name__)
 
-SENSOR_DESCRIPTIONS: Final[tuple[SensorEntityDescription, ...]] = (
-    SensorEntityDescription(
+
+@dataclass
+class HyundaiKiaSensorEntityDescription(SensorEntityDescription):
+    """A class that describes custom sensor entities."""
+
+    exists: Callable[[Vehicle], bool] | None = None
+
+
+def _is_electrified(vehicle: Vehicle) -> bool:
+    """Return True for BEV and PHEV vehicles."""
+    return vehicle.engine_type in (ENGINE_TYPES.EV, ENGINE_TYPES.PHEV)
+
+
+SENSOR_DESCRIPTIONS: Final[tuple[HyundaiKiaSensorEntityDescription, ...]] = (
+    HyundaiKiaSensorEntityDescription(
         key="_total_driving_range",
         translation_key="total_driving_range",
         icon="mdi:road-variant",
@@ -40,7 +55,7 @@ SENSOR_DESCRIPTIONS: Final[tuple[SensorEntityDescription, ...]] = (
         native_unit_of_measurement=DYNAMIC_UNIT,
         state_class=SensorStateClass.MEASUREMENT,
     ),
-    SensorEntityDescription(
+    HyundaiKiaSensorEntityDescription(
         key="_odometer",
         translation_key="odometer",
         icon="mdi:speedometer",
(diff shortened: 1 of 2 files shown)


custom_components/kia_uvo/sensor.py
```diff
@@ -479,6 +479,23 @@ async def async_setup_entry(
                     vehicle.engine_type in (ENGINE_TYPES.EV, ENGINE_TYPES.PHEV)
                     or vehicle.ev_charging_power is not None
                 )
+            elif description.key in (
```
cdnninja: I think it is time to define an exists function on the descriptions and move it all up.  This is getting very large.

blka: Done in 837ba8d. `HyundaiKiaSensorEntityDescription` gains `exists: Callable[[Vehicle], bool] | None = None` — the pattern binary_sensor.py already uses for `is_on` — and `async_setup_entry` is down to a single check:

```python
create = (
    description.exists(vehicle)
    if description.exists is not None
    else getattr(vehicle, description.key, None) is not None
)
```

Gate semantics ported 1:1 from the elif branches, comments moved with their gates, `_is_electrified()` dedups the EV/PHEV check across the three electrified gates. 66 tests pass, ruff 0.16.4 clean.

PR #1856, opened 2026-08-30
fix(sensor): always create geocoded location sensor

### LINKED ISSUE #1844: Canada: Location and Geocoded location sensors unavailable again
### Region

CA

### Brand of Car

Kia

3.10.1

### Describe the Bug

This issue seems to be back, for me at least, as of today. v3.10.1. Last time this happened it was gone for weeks and then came back.

Debug logging doesn't provide any more information.  That's just from enabling debug logging from the HACS integration.  

```
Logger: hyundai_kia_connect_api.KiaUvoApiCA
Source: /usr/local/lib/python3.14/site-packages/hyundai_kia_connect_api/KiaUvoApiCA.py:932
First occurred: 2:15:05 PM (1 occurrence)
Last logged: 2:15:05 PM

hyundai_kia_connect_api - Get vehicle location failed
```

### Diagnostics dump (if available)

_No response_

### To Reproduce

Reboot HASS, reload the integration or force a refresh.

### Expected Behavior

Location and geocoded location sensors should show location.

### Screenshots

_No response_

### Additional Context

_No response_

blka:
## Summary

The geocoded-location sensor is created at setup only when a geocode is already known (`sensor.py` creation gate: `getattr(vehicle, key, None) is not None`). The geocode lives in memory next to the vehicle coordinates: after a Home Assistant restart or integration reload it is `None` until a poll fetches coordinates and the geocode lookup succeeds.

While it is `None` at setup, the entity isn't yielded, and HA marks it **"no longer provided"** — it shows unavailable and never returns until the next reload, even after location recovers on a later poll. Combined with a flaky Find-My-Car response this is kia_uvo #1844: *"gone for weeks and then came back"* — it came back on a restart, not when the car moved.

## Changes

- `custom_components/kia_uvo/sensor.py`: always create the `_geocode_name` sensor. A `None` value renders as HA `unknown` and the real value arrives on the first successful geocode, without requiring a reload.

## Scope

- Same pattern as the 12V SoC sensor (#1803): transient value → don't gate creation on it.
- The device_tracker already creates unconditionally and turns available on the first poll with coordinates — no change needed there.
- The location-recovery itself (fetch after restart when the car is parked) is handled on the library side by Hyundai-Kia-Connect/hyundai_kia_connect_api#1294; this PR covers the entity that would otherwise stay stale-unavailable even after recovery.

Fixes Hyundai-Kia-Connect/kia_uvo#1844

### DIFF
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -538,6 +538,17 @@ async def async_setup_entry(
     for vehicle_id in coordinator.vehicle_manager.vehicles:
         vehicle: Vehicle = coordinator.vehicle_manager.vehicles[vehicle_id]
         for description in SENSOR_DESCRIPTIONS:
+            if (
+                description.key == "_geocode_name"
+                and not coordinator.vehicle_manager.geocode_api_enable
+            ):
+                # Created only when the geolocation option is on: without it
+                # the library never runs the geocode lookup, so the sensor
+                # would sit `unknown` forever. With the option on, a None at
+                # setup must not gate creation — see the `_geocode_name`
+                # description above. Gated here, not in `exists`, because the
+                # option is coordinator state.
+                continue
             create = (
                 description.exists(vehicle)
                 if description.exists is not None
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -281,6 +281,13 @@ SENSOR_DESCRIPTIONS: Final[tuple[HyundaiKiaSensorEntityDescription, ...]] = (
         key="_geocode_name",
         translation_key="geocode_name",
         icon="mdi:map",
+        # The geocoded address is transient — it lives in memory and is None
+        # right after a restart/reload until a poll with coordinates
+        # succeeds. Don't gate creation on it: a None at setup means the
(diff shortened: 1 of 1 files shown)


cdnninja:
Shouldn't this be configuration based?  Users may not have it enabled.

blka:
Yes — good catch. With the option disabled the library never runs the geocode lookup (`VehicleManager` only calls `update_geocoded_location` when `geocode_api_enable` is set), so the always-created sensor would sit `unknown` forever. Creation is now gated on the option (28ab799); with it enabled the sensor is always created and recovers from `unknown` on the next successful poll — which is the #1844 case.

blka:
check https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/1847 first, if ok I will rebase

cdnninja:
Ready for rebase.

blka:
Rebased onto master after #1847. The fix was re-expressed for the new exists-on-descriptions mechanism: `exists=lambda _: True` on the `_geocode_name` description (same pattern as `car_battery_percentage`), and the geolocation-option gate moved into `async_setup_entry` — the option is coordinator state, so it can't live in a module-level description lambda. Semantics unchanged: option off → no entity; option on → always created, a None renders `unknown` until the first successful geocode (kia_uvo #1844). CI green.

PR #1864, opened 2026-09-03
chore: add BR region to issue template and label workflow

blka:
## Changes

- **Bug report region dropdown**: added `BR` (Brazil), which the integration has supported since region 8 was added to `const.py` — the dropdown only offered EU/CA/USA/AUS/China/India/NZ.
- **`label-issue.yml`**: region regex updated to also match `BR`, so issues selecting BR get labeled.
- **Version placeholder**: refreshed from the stale `0.4.1` to the current `3.11.0` (per `manifest.json`).
- **Feature request template** (second commit): replaced the legacy front-matter `feature_request.md` with a YAML form matching `bug_report.yml` — region (incl. BR) and brand dropdowns, version input, structured textareas.

## Notes for review

While auditing the templates I found a few label-hygiene items in the repo settings that a PR can't fix — happy to help with any of them:

1. The repo has no `Brazil`, `China`, `India`, or `NZ` labels yet, so issues from those regions currently get an auto-created default-gray label (if any). Consider pre-creating them.
2. There are duplicate region labels: `AUS`/`Australia` and `CA`/`Canada` — the dropdown uses the short forms, so the long-form ones appear unused.
3. ~~`feature_request.md` is still a legacy front-matter template~~ — now converted to a YAML form in this PR.

/cc @cdnninja — for the two remaining notes above, let me know if you agree and I'll follow up (or feel free to handle them directly).

All changes are `.github/` only; validated with pre-commit (prettier/yamllint).


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -20,6 +20,7 @@ body:
         - China
         - India
         - NZ
+        - BR
```
cdnninja: I am thinking for new we start using the full word? I added Brazil as a tag.   Ideally I think we move them all to the full name.   Since for example EU is no longer accurate we can move that to Europe.

blka: Done — all region options (both templates) and the labeler regex now use full names: Europe, Canada, United States, Australia, China, India, New Zealand, Brazil. Rebased onto master as well. Note: I don't have permission to create the missing repo labels — `Europe`, `United States`, `New Zealand` don't exist yet (Brazil/Australia/Canada/China/India do). The workflow's `addLabels` will auto-create them in gray if they're absent, but you may want to create them with proper colors. The old short labels (`EU`, `CA`, `USA`, `AUS`) can be retired afterwards.

PR #1873, opened 2026-09-06
fix: stabilize vehicle identity and retain odometer sensor

### LINKED ISSUE #1852: Duplicate vehicle device created and missing sensors after upgrading to paid Bluelink Plus (Kona EV 2021)
### Region

EU

### Brand of Car

Hyundai

2.36.3 (I have an old HA, 2024.12 and I am unable to update it, kia_uva integration is still working)

### Description
I searched around but couldn't find a topic covering this specific problem. I'm quite new to this, so I'm not sure if this is a bug or intended behavior, but here is what I'm experiencing:

After upgrading my Hyundai Bluelink account from the 5-year free plan to the paid **Bluelink Plus** tier (Hyundai Kona Electric 2021), reloading the integration resulted in two vehicle devices appearing instead of one:

1. **Orphaned Device:** Contains old entities marked with *"This entity is no longer being provided by the kia_uvo integration..."*
2. **New Device:** Active, but several telemetry and energy sensors previously available under the free tier are missing.

### Missing Entities
The following sensors are missing from the newly generated device:
* `sensor.kona_average_energy_consumption`
* `sensor.kona_daily_driving_stats`
* `sensor.kona_odometer`
* `sensor.kona_todays_daily_driving_stats`
* `sensor.kona_total_energy_consumption`
* `sensor.kona_total_energy_regeneration`

### Environment Information
* **Vehicle:** Hyundai Kona Electric (MY2021)
* **Subscription:** Bluelink Plus (Paid)
* **Integration:** `kia_uvo` / Hyundai-Kia-Connect - 2.36.3 (I have an old HA, 2024.12 and I am unable to update it, kia_uva integration is still working)

### Additional Context / Questions
Is there a way to force a full re-sync to recover these entities, or did Hyundai change/deprecate payload endpoints for the Bluelink Plus tier on 2021 models?

_No response_

### Diagnostics dump (if available)

_No response_

### To Reproduce

-

### Expected Behavior

* A single active vehicle device in Home Assistant without ghost entities.
* Retention of all driving, odometer, and energy consumption sensors supported by the Bluelink API.

### Screenshots

_No response_

### Additional Context

_No response_

Brahma-Agni:
## Summary

Addresses the verified parts of #1852:

- Adds the VIN as a stable secondary device identifier while retaining Hyundai's existing `vehicle.id` identifier.
- Keeps the odometer entity registered when its value is temporarily absent during setup.

The proposed forced creation of EV energy and daily-driving-stat entities has been removed following maintainer feedback because support is not confirmed across all models, regions, and subscription levels.

## Device identity

The integration previously identified a Home Assistant device only by Hyundai's backend-provided `vehicle.id`. The device now advertises both that existing identifier and, when available, `vin:<VIN>`:

```python
identifiers = {(DOMAIN, self.vehicle.id)}
if vin := self.vehicle.VIN:
    identifiers.add((DOMAIN, f"vin:{vin}"))
```

Keeping `vehicle.id` preserves matching for existing installations. The VIN provides a stable identity if Hyundai later changes its internal vehicle ID.

This was also validated in place on a real EU Hyundai/Home Assistant installation: reloading the PR branch kept the same Home Assistant device and all existing entities, added the VIN identifier, and created no duplicate device or entity churn.

## Odometer availability

Odometer is a stable vehicle capability, but its value may be absent in the initial payload. The entity is now always registered and reports `unknown` until a later refresh supplies a value, instead of being marked as no longer provided.

## Tests

Regression coverage verifies that:

- Vehicle records with different backend IDs and the same VIN share the stable VIN identifier.
- The odometer entity is created when its initial value is `None`.

Validation completed successfully:

- Ruff formatting and static analysis
- Diff validation
- Complete test suite: **73 passed**

## Scope

This change does not force creation of EV energy or daily-driving-stat entities, and it does not automatically merge duplicate devices that already existed before the VIN identifier was registered.

### DIFF
--- a/custom_components/kia_uvo/entity.py
+++ b/custom_components/kia_uvo/entity.py
@@ -27,8 +27,13 @@ class HyundaiKiaConnectEntity(
     @property
     def device_info(self) -> DeviceInfo:
         """Return device information to use for this entity."""
+        identifiers = {(DOMAIN, self.vehicle.id)}
+        if vin := self.vehicle.VIN:
+            # Hyundai may replace its internal vehicle id after an account or
+            # subscription migration. The VIN keeps HA attached to the same car.
+            identifiers.add((DOMAIN, f"vin:{vin}"))
         return DeviceInfo(
-            identifiers={(DOMAIN, self.vehicle.id)},
+            identifiers=identifiers,
             manufacturer=f"{BRANDS[self.coordinator.vehicle_manager.brand]} {REGIONS[self.coordinator.vehicle_manager.region]}",
             model=self.vehicle.model,
             name=self.vehicle.name,
--- a/custom_components/kia_uvo/sensor.py
+++ b/custom_components/kia_uvo/sensor.py
@@ -62,6 +62,9 @@ SENSOR_DESCRIPTIONS: Final[tuple[HyundaiKiaSensorEntityDescription, ...]] = (
         native_unit_of_measurement=DYNAMIC_UNIT,
         device_class=SensorDeviceClass.DISTANCE,
         state_class=SensorStateClass.TOTAL_INCREASING,
+        # Odometer is a stable vehicle capability, but its value can be absent
+        # during setup. Keep the entity so a later refresh can populate it.
+        exists=lambda _: True,
     ),
     HyundaiKiaSensorEntityDescription(
         key="_last_service_distance",
@@ -236,6 +239,9 @@ SENSOR_DESCRIPTIONS: Final[tuple[HyundaiKiaSensorEntityDescription, ...]] = (
         native_unit_of_measurement=UnitOfEnergy.WATT_HOUR,
         device_class=SensorDeviceClass.ENERGY,
         state_class=SensorStateClass.TOTAL,
+        exists=lambda vehicle: (
+            _is_electrified(vehicle) or vehicle.total_power_consumed is not None
+        ),
     ),
     HyundaiKiaSensorEntityDescription(
         key="total_power_regenerated",
@@ -244,6 +250,9 @@ SENSOR_DESCRIPTIONS: Final[tuple[HyundaiKiaSensorEntityDescription, ...]] = (
         native_unit_of_measurement=UnitOfEnergy.WATT_HOUR,
         device_class=SensorDeviceClass.ENERGY,
         state_class=SensorStateClass.TOTAL,
+        exists=lambda vehicle: (
+            _is_electrified(vehicle) or vehicle.total_power_regenerated is not None
+        ),
     ),
     # Need to remove km hard coding.  Underlying API needs this fixed first.  EU always does KM.
     HyundaiKiaSensorEntityDescription(
@@ -252,6 +261,9 @@ SENSOR_DESCRIPTIONS: Final[tuple[HyundaiKiaSensorEntityDescription, ...]] = (
         icon="mdi:car-electric",
         state_class=SensorStateClass.MEASUREMENT,
         native_unit_of_measurement=f"{UnitOfEnergy.WATT_HOUR}/km",
+        exists=lambda vehicle: (
+            _is_electrified(vehicle) or vehicle.power_consumption_30d is not None
+        ),
     ),
     HyundaiKiaSensorEntityDescription(
         key="front_left_seat_status",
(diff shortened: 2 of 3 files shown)


cdnninja:
To confirm has the ID side been tested real world? 

As for the exists side, do we know that these features are support by all models, regions and subscription levels for PHEV and EV?  Do we know why the data wasn't returned?

Brahma-Agni:
Thanks for the review, @cdnninja.

The VIN identifier change has only been unit-tested and has not yet been verified on a live vehicle/account migration.

For the sensor changes, we do not yet know whether every EV/PHEV model, region, and subscription supports these fields. Issue #1852 does not include diagnostics, so we cannot confirm why the data was missing.

I’ll request diagnostics from the reporter and wait for real-world confirmation before proceeding.

cdnninja:
I am less worried on the VIN change for migration but rather does it impact people who don't migrate accounts?  So testing in place is key.   I don't have one of these cars anymore to confirm HA behavior.

Brahma-Agni:
That makes sense. For users who do not migrate accounts, the existing vehicle.id identifier is still retained, so Home Assistant should continue matching the same device exactly as before. The VIN is only added as a secondary identifier to make that device resilient if the backend ID changes later; it does not replace the current identifier.

The unit tests confirm that the existing identifier remains present and that records with the same VIN share a stable identifier. The full test suite also passes, so I do not expect the change to affect existing installations that have not migrated.

@cdnninja can I add a Home Assistant device-registry test that starts with an existing device, adds the VIN identifier during reload, and verifies that no additional device is created. Please let me know if you would prefer that test or any other change to the implementation.

blka:
@cdnninja Live data for exactly this question, from a real-world in-place install (EU Hyundai, SANTA FE HEV) that recently went through the scenario behind #1852: the Bluelink subscription expired, the vehicle dropped to the basic plan, and was then re-upgraded to Bluelink Pro.

Checked today, before this PR is installed, from the device/entity registry plus a fresh API login:

- the backend `vehicle.id` returned by the API is the same id the device was originally registered with — unchanged by the subscription migration
- exactly one `kia_uvo` device exists (no duplicate), with 100 entities on it, none disabled
- the device registry entry has not been touched since April; the config entry loads fine

So on this real migration path the backend id stayed stable and an unpatched install survived it without any device duplication. One data point — it does not prove ids never change on other paths (account migration, VIN reassignment), which is where your PR change helps.

Next I'll run this PR branch in-place on my test instance and confirm the existing device gains the `vin:` identifier on reload without a duplicate device or entity churn.

blka:
Follow-up on the in-place testing question — I ran the PR branch (65bbe37) on a Docker HA 2026.7.4 test instance against a real EU Hyundai account (SANTA FE HEV), reusing the pre-existing config entry and device registry that were created before the PR, so the reload exercised exactly the in-place path @cdnninja asked about.

**Result: merge works in-place, no duplication, no impact on non-migrated accounts.**

Verified on the running instance:

- Before reload: 1 device, identifiers `[['kia_uvo', '<vehicle_id>']]`, 97 entities.
- After reload with the PR loaded: still **1 device** (same `device_id`), identifiers now `[['kia_uvo', 'vin:<VIN>'], ['kia_uvo', '<vehicle_id>']]`, all 97 existing entities still bound to the same device_id. `modified_at` is the only other change.
- No new device was created and no entity re-registered to a second device — the registry lookup matches the existing device via the vehicle.id identifier and `device_registry.async_get_or_create` merges the `vin:` identifier into it (confirmed by reading `.storage/core.device_registry` before/after).
- 3 additional binary_sensor entities registered on first load (existing feature keys this test instance hadn't registered before) — unrelated to the device identity change, no second device involved.

For reference on the migration scenario itself (posted earlier in this thread): on a real account that went through a subscription change, the EU API kept the same vehicle.id, so people who don't migrate accounts see no registry change at all beyond the identifier being added on the first reload after upgrade — which is exactly what this test confirms happens cleanly.

cdnninja:
I am good with the VIN change and the odometer change.  The EV ones without knowing the reason I am not sure we should force them. I expect forcing to cause other tickets for cars that don't support it.

cdnninja:
Any update on the above feedback?

Brahma-Agni:
Thanks for the follow-up, @cdnninja. I have addressed the feedback in commit 25d92a4.

The forced creation of the EV energy and daily-driving-stat entities has been removed. The PR now contains only the VIN identifier and odometer changes you approved, and the regression test has been narrowed to match that scope.

I also updated the PR title and description. Ruff checks pass and the complete test suite passes with 73 tests.

PR #1895, opened 2026-09-17
feat: runtime library version override from the options flow

blka:
## Summary

Adds an **advanced, opt-in library version override** to the integration options flow. It lets a user run a different `hyundai_kia_connect_api` version — without editing `manifest.json` and without reinstalling the integration — from **Settings → Devices & Services → kia_uvo → Configure (options)**.

This is useful when testing a fix from an upstream PR of the API library, pinning a known-good older version, or loading a local build, directly from the Home Assistant UI.

## How it works

A single free-text field, `library_override`, is added to the options schema (empty = the manifest-pinned version, so nothing changes for existing users). On entry setup, before the coordinator is created, the integration interprets the value:

- **Empty** — no-op; the manifest-pinned version applies (Home Assistant core already reinstalls the manifest requirement on every start).
- **A released version** (`4.28.0` or `==4.28.0`) — no-op when already installed; otherwise the manifest requirement is rebuilt with the requested version (manifest extras such as `[image]` are kept) and installed.
- **Any pip requirement verbatim** (git+https, file://, …) — installed with `--force-reinstall --no-deps` (pip reports a spec resolving to the installed version as satisfied, so the swap must be forced; `--no-deps` keeps dependency versions untouched).

After a version change, the loaded library and integration modules are purged from `sys.modules` and the loader cache is reset, then `ConfigEntryNotReady` is raised so the setup retry imports the fresh library. A failed install raises `ConfigEntryError` (entry shows the error in the UI) instead of retrying a never-succeeding install forever; a PR page URL pasted from the browser instead of the `@refs/pull/N/head` git ref is detected and reported with the corrected form.

The library is a single site-packages install shared by all kia_uvo entries, so the first entry that applies an override wins and a second entry requesting a different override only logs a warning; the entry that applied it can change or clear its own override freely. Clearing the field restores the manifest-pinned version immediately (same-session reload) and logs it.

The diagnostics dump already reports `library_version` and now also `library_source` (from PEP 610 `direct_url.json`), which shows the actual install source of the active library (git URL + ref + commit for a PR/branch override, `file://` for a local install, absent for a PyPI pin).

## Changes

- `const.py` — options key, `hass.data` marker/lock keys, package name constant
- `__init__.py` — override install/restore logic, called at the top of `async_setup_entry`
- `config_flow.py` — the new options field
- `diagnostics.py` — `library_source` from `direct_url.json`
- `strings.json` + `translations/en.json` / `pl.json` — new field label and description
- `README.md` — "Runtime library version override (advanced)" section

## Testing

All live-tested against a Home Assistant 2026.7.4 container with a real EU account:

- downgrade `4.31.0` → `4.28.0` and back via the manifest rebuild (re-import proof: entities re-registered against the fresh library)
- PR head override `@refs/pull/1303/head` (diagnostics dump shows the git URL + ref + commit from `direct_url.json`)
- fork branch, commit SHA, and `file://` specs
- no-op when the requested version is already installed; empty field restores the pin
- fail-fast on a nonexistent ref (`couldn't find remote ref`) — single attempt, clear error, no retry loop
- same-session override change and clear-restore on the entry that applied it

Lint (`ruff` 0.16.7, all tracked files) and `mypy --strict` on the changed files pass.

## Caveats (documented in the README)

- A git/URL requirement is reinstalled once per Home Assistant start (there is no version to compare) with `--no-deps`.
- A verbatim pip requirement does not inherit the manifest extras.
- With several kia_uvo entries, the first override applied wins for the session.


cdnninja:
What problem are we trying to solve? For occasional testing I think this is feature bloat.

blka:
The driver is the EU Kia GSPA transition: fixes land in `hyundai_kia_connect_api` (PRs or point releases) faster than kia_uvo can ship releases, so users on a single kia_uvo version can test or apply the lib fix — a PR head or a pinned version — from the options UI instead of hand-editing `manifest.json`, and the diagnostics dump reports the actually installed version and source. The field is opt-in; empty keeps today's behavior. If you'd still rather not carry the pip-in-runtime surface, happy to close this — it started as a local testing tool.

PR #1897, opened 2026-09-19
fix(sensor): retain /drvhistory entities across data-less update cycles

### LINKED ISSUE #1896: EV efficiency entities disappear after one update cycle without /drvhistory data; keep them as `unavailable` instead
**Reported via Hyundai-Kia-Connect/hyundai_kia_connect_api#1315** (Ioniq 5, EU): the five EV efficiency entities — `total_power_consumed`, `power_consumption_30d`, `total_power_regenerated`, `DailyDrivingStatsEntity`, `TodaysDailyDrivingStatsEntity` — all disappeared together after an HA restart and came back after the next one.

## Root cause

All five are fed by a single endpoint: `POST /drvhistory` (`KiaUvoApiEU._get_driving_info()` in the library, two calls per update cycle: `periodTarget: 1` (alltime) and `periodTarget: 0` (30d)).

In `sensor.py` (around lines 560–578), the entities are only **created** when the values are present at coordinator setup. One update cycle where `/drvhistory` returns no data (e.g. the car hasn't been driven recently — the reporter hadn't driven since Monday) leaves the values as `None`/empty, so after a restart the entities are never added. They only reappear after a subsequent restart once the endpoint returns data again.

## Proposed change

Keep the entities alive once created and show `unavailable` when the values are `None`, instead of dropping them after a single cycle without data.

Secondary improvement: when `/drvhistory` returns an empty response without an error code, `_get_driving_info()` currently sets the fields to `None` silently. A debug/info log line there would make this failure mode visible (the error path via `_check_response_for_errors` is already logged as `Failed to parse driving info`).

## Sequencing

To avoid conflicts, implement this after the mypy-strict PR (#1893) is merged.

blka:
## Summary

The five EV efficiency entities fed by `POST /drvhistory` were only created when their values were present at HA setup (the `exists`/data gating in `sensor.py` plus the `vehicle.daily_stats` check). One update cycle without data after a restart (an idle car, or an empty endpoint response) dropped all of them at once, and they only came back after the next reload. Observed end-to-end in Hyundai-Kia-Connect/hyundai_kia_connect_api#1315: entities gone after a restart, back after the next one.

Keep them for electrified vehicles in every region whose library implementation polls `/drvhistory`, and report `unknown` when the endpoint returns no data — the same pattern as the `_geocode_name` sensor (#1844) and the 12 V SoC sensor (#1803).

## Changes

- `sensor.py`:
  - The three description-based `/drvhistory` sensors (`total_power_consumed`, `total_power_regenerated`, `power_consumption_30d`) are gated on **region × engine type** instead of data presence: EU/India impls poll for EV/PHEV, China for EV only, Australia (also used for NZ Kia) for EV/PHEV; CA/USA/BR impls never poll. This mirrors `VehicleManager.get_implementation_by_region_brand` + the gates in each implementation (`KiaUvoApiEU`, `KiaUvoApiIN`, `KiaUvoApiCN`, `KiaUvoApiAU`).
  - `total_power_regenerated` stays EU/India-only: the AU and CN implementations never parse `regenPwr` into it, so an always-created entity there would sit `unknown` forever.
  - `DailyDrivingStatsEntity` + `TodaysDailyDrivingStatsEntity` are always created in /drvhistory-polling regions; an empty/`None` `daily_stats` reports `unknown` (not a misleading 0 days), and both attribute loops are None-safe (`or []`).
  - The region gate reads `REGIONS.get(region, "")` — `region` is the int index into `REGIONS` (config_flow stores ints; same shape `entity.py` already relies on for the manufacturer string).
- No entity keys or translations changed — same five entities, same `unique_id`s; entities missing today in an affected setup simply come back after this update + reload.

## Deviation from #1896 wording

The ticket said `unavailable`; this implements `unknown`. Empty `/drvhistory` data is not a coordinator/device failure — it is a valid state for the data source, and HA semantics reserve `unavailable` for entity/coordinator availability failures (`last_update_success`). Happy to switch if maintainers prefer `available = False`.

## Verification

- `mypy --strict --ignore-missing-imports custom_components/kia_uvo/`: 0 errors in 20 files (HA 2026.9.2 + lib from `manifest.json` pin).
- `ruff check` + `ruff format --check` (0.16.7, pre-commit pin): clean; pre-commit hooks pass on changed files.
- pytest: **81 passed** (71 existing + 10 new in `tests/test_driving_info_sensor_creation.py`).
- New tests cover: EU EV/PHEV without data (entities kept, `unknown`), EU EV with data (values present), India EV (regenerated included), AU EV and CN EV (no regenerated), CN PHEV (not created), USA EV (not created), EU ICE (not created), and the int region-id → `REGIONS` mapping used by the tests.
- `compileall` on the package: clean.

## Scope

`sensor.py` + one new test file; single commit. No lib changes — follow-ups noted in Hyundai-Kia-Connect/hyundai_kia_connect_api#1315 (a log line when `/drvhistory` returns an empty response, and stale "Europe feature only" comments in `Vehicle.py` — AU/CN/IN populate these fields too).

Closes #1896


cdnninja:
Our goal should be to avoid doing region specific logic on the ha side. I would rather do a "supports" entity on the vehicle. 

Another option since this seems to be a thing, could we do on demand adding the entity if data returns?  It would universally fix this. Thoughts?

blka:
Agreed on the goal — region literals in `sensor.py` are the weakest part of this PR. Both suggestions make sense to me; I'd sequence them rather than pick one:

**`supports` flags on the vehicle — this is the better fix, and I'd rework the PR to it.** The region×engine matrix here is a proxy for a capability the library already decides: each impl knows whether it polls `/drvhistory` and whether it parses `regenPwr` into `total_power_regenerated` (AU/CN don't — that's why `total_power_regenerated` is EU/IN-only in the PR; an always-created entity there would sit `unknown` forever). So two flags set by the impl, e.g. `vehicle.supports_driving_history` + `vehicle.supports_power_regeneration`, would let `sensor.py` gate on the vehicle alone and drop every region literal. Same shape as existing capability fields (`ccu_ccs2_protocol_support`, `ccs2_protocol_support`), and it composes: the gate becomes "flag present → entity exists", regardless of how the impl derives it (engine type, region, future server-side capability data). Happy to rework this PR onto the flags — the diff is small since the impls already branch exactly where the flags would be set.

**On-demand creation when data returns — works, but I'd keep it as a fallback, not the primary fix.** The coordinator would need to hold the `async_add_entities` callback and create entities on new keys mid-cycle, which is more plumbing than the flag approach; entities would also flicker `unknown`-less — before first data they simply don't exist, so dashboards show gaps until data arrives, and a car that never returns data (rate-limited, older model) silently has no entities at all while an identical car elsewhere has them. And when data appears after being absent, the entity registers fresh and starts `unknown` anyway — so the end state equals the flag approach, just with churn in the registry and gaps in long-term statistics. Where it *does* win: capabilities the library can't know at setup. For `/drvhistory` the library knows at setup, which is why I'd keep creation unconditional and values `unknown` when data is absent.

One detail to settle if we go the flags route: where they live. Deriving from `engine_type` in `sensor.py` would still leave region knowledge on the HA side; setting the flag inside each impl keeps the knowledge in one place — that's where I'd put it.

Issue #1888, opened 2026-09-12
Authentication expired for Hyundai Europe


andyaiprof-jpg:
### Region

Europe

### Brand of Car

Hyundai

### Version of the Integration

3.12.0

### Describe the Bug

HA → Hyundai API erreichbar ✅ → Device ID erstellt ✅ → OAuth authorize ❌ → AuthenticationError

### Debug Logs (if an error occurred)

Get Device ID response:
'retCode': 'S',
'resCode': '0000'
Get cookies request:
https://prd.eu-ccapi.hyundai.com:8080/api/v1/user/oauth2/authorize...

### To Reproduce

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/fc404c0b-3deb-45ca-81ab-6703547fc47b" />

### Expected Behavior

Reconnect works

### Screenshots

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/3d39543a-c272-4c91-ad46-b426940fb506" />

gertvh:
Same failure here on EU Hyundai (`kia_uvo` 3.12.0 / `hyundai_kia_connect_api` 4.29.1): device ID registration succeeds, then it stops right at the cookies/authorize step. I traced it against the live servers from the HA host, and it looks like a server-side change rather than credentials or config.

Hyundai has moved the EU login onto a new identity system (the RSA cert now comes back with `kid: HMGID2_CIPHER_KEY1`), and the endpoints the library relies on now redirect into a new React SPA served under `/web/v1/user/` instead of accepting the old flow.

Where the current flow breaks:

- `GET https://prd.eu-ccapi.hyundai.com:8080/api/v1/user/oauth2/authorize?...` (the `_get_cookies()` call, which is the last line in my debug log too) → 302 → `/web/v1/user/401`. `_get_cookies()` doesn't inspect the response, so it collects cookies from that SPA page and continues.
- `GET https://idpconnect-eu.hyundai.com/auth/api/v2/user/oauth2/authorize?client_id=<oneapp>&...` → 302 → `https://prd.eu-ccapi.hyundai.com:8080/web/v1/user/authorize?...&next_uri=<the idpconnect authorize url>`. The OneApp/CCI authorize now hands off to the same SPA.
- `POST https://idpconnect-eu.hyundai.com/auth/account/signin` → 302 back to the authorize URL instead of returning a code, so the old RSA form-post no longer accepts credentials.
[...]

Issue #1322, opened 2025-09-21
Unable to login to app


GogartyIT:
Please check Services, Known Bug / Issues and Troubleshooting over here first: https://github.com/fuatakgun/kia_uvo/blob/master/README.md
**Region and Brand of car
Kona, EU

**Version of the integration**
Latest, HACS version, e.g. 2.43.9

**Describe the bug**
Umable to login
Network Tests i did,

curl -v --connect-timeout 5 https://prd.eu-ccapi.hyundai.com:8080/
→ Connection timed out from:

Home Assistant host (wired broadband)

Laptop on mobile hotspot

DNS resolution works: prd.eu-ccapi.hyundai.com -> 103.187.35.235

Port 443 responds with a TLS handshake (curl -v https://prd.eu-ccapi.hyundai.com:443/ shows a 400 error).

**Debug logs if an error occurred**

This error originated from a custom integration.

Logger: custom_components.kia_uvo.config_flow
Source: custom_components/kia_uvo/config_flow.py:148
integration: kia_uvo (documentation, issues)
First occurred: 20 September 2025 at 16:42:44 (2 occurrences)
Last logged: 20 September 2025 at 16:48:45

Unexpected exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/urllib3/connection.py", line 198, in _new_conn
    sock = connection.create_connection(
        (self._dns_host, self.port),
    ...<2 lines>...
        socket_options=self.socket_options,
    )
  File "/usr/local/lib/python3.13/site-packages/urllib3/util/connection.py", line 85, in create_connection
    raise err
  File "/usr/local/lib/python3.13/site-packages/urllib3/util/connection.py", line 73, in create_connection
    sock.connect(sa)
    ~~~~~~~~~~~~^^^^
TimeoutError: [Errno 110] Operation timed out

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/urllib3/connectionpool.py", line 787, in urlopen
    response = self._make_request(
        conn,
    ...<10 lines>...
        **response_kw,
    )
  File "/usr/local/lib/python3.13/site-packages/urllib3/connectionpool.py", line 488, in _make_request
    raise new_e
  File "/usr/local/lib/python3.13/site-packages/urllib3/connectionpool.py", line 464, in _make_request
    self._validate_conn(conn)
[...]

PR #1899, opened 2026-09-20
fix(config): persist fresh token on reconfigure with new credentials

blka:
## Summary

The credentials step of **reconfigure** already performs a full login with the new credentials (`validate_input` → `VehicleManager.login()`), but the token from that login was discarded: `data_updates` merged only the new fields, so the entry kept the stale persisted token. On reload the coordinator refreshed from the stale token first and only recovered after a failed refresh attempt (with a `CCI token refresh failed, falling back to full login` warning in the log).

The OTP completion path already persists the fresh token (`config_flow.py` OTP branch); the plain-password reconfigure path now does the same. The reauth path is unchanged (full data replace, intentionally drops the token).

## Dependency on the library PR

**Depends on Hyundai-Kia-Connect/hyundai_kia_connect_api#1322** — the library-side recovery that retries login with the configured credentials when the persisted token wedged on the legacy 48-char refresh-token password (the wedge behind the integration failures in Hyundai-Kia-Connect/kia_uvo#1888). That PR fixes the wedge itself; this change makes the recovery unnecessary for accounts reconfigured with fresh credentials by reusing the token the validation login already obtained (and removes the confusing refresh-failure warning on every reconfigure). Merge order: library PR first; this one still functions without it (the stale-token refresh path remains as before), it just wastes one refresh attempt until #1322 lands and the manifest pins a released version containing it.

## Validation

- `ruff check` / `ruff format --check` on `custom_components/kia_uvo/`: clean
- No entity or behavior change outside the config flow; hassfest runs in CI


custom_components/kia_uvo/config_flow.py
```diff
@@ -244,9 +244,16 @@ async def async_step_credentials_password(
 
                     return await self.async_step_select_otp_method()
                 if self._is_reconfigure:
+                    # validate_input performed a full login with the new
```
cdnninja: Not sure we need this comment?

blka: Removed — the persistence reads clearly enough from the diff.

blka:
@cdnninja your comment on the config-flow prose is addressed — comment removed (bad6a9d), thread resolved. Ready for a re-review whenever you have a moment.

